Issue #17, opened 2023-01-12
Micro manager should be able to control micro simulations written in C++


IshaanDesai:
Currently the Micro Manager is constructed in a way that is can only control micro simulations written in Python. The user gives a path to a micro simulation script written in Python, in the manager configuration file. However it can be the case that a user has a micro simulation script written in some other language like C++, C of even Fortran. For the time being the manager can be updated in a way that scripts written in C and C++ can be used. Both C and C++ scripts cannot be used directly, and in most cases will be built into a shared library (`.so`) file. The Micro Manager needs to take the shared library file and run the appropriate functions.

For micro simulations written in C++, [pybind11](https://github.com/pybind/pybind11) was the tool used in https://github.com/precice/micro-manager/pull/22.

fsimonis:
A possibly easier (less-invasive) solution is to restructure the C++ code into a shared library, then dynamically load and call it. The user has to restructure the code anyhow.

[Here is a writeup](https://realpython.com/python-bindings-overview/) on how to do this.

IshaanDesai:
An example of what @fsimonis is describing is:

* The C++ file: https://github.com/precice/aste/blob/1.1.0/src/metisAPI.cpp
* Loading the shared library file and using it in a Python script: https://github.com/precice/aste/blob/53a4f0caf6e43ca787c7eff148bb51c19cfe81cc/src/partition_mesh.py#L172

fsimonis:
Quick drawing of this
![image](https://user-images.githubusercontent.com/13552216/212309723-3e45f056-a298-48e3-b90f-fa26e361cab8.png)

erikscheurer:
At the moment, I am looking into the shared library approach, which seems doable. Is there a way to pass python dictionaries between the languages? This is how the micromanager handles the information passing to the micro solver at the moment. 
They can contain different amounts of keys-value pairs and can contain either floating points or lists of floats, from what I can tell. So declaring the input to the c++ `solve` function as `std::map<std::string, std::vector<float>> macro_data` only works if we pass the scalar data also as vectors, but thats probably to much overhead?

erikscheurer:
The current approach would be to split up the dictionary in the python wrapper and create new c compatible variables. Meaning we would iterate through the `macro_data` dictionary and differentiate between floats and ndarrays to create pointers to this data. Then pass these new pointers to the c++ side with
[...]

PR #20, opened 2023-01-26
Add C++ bindings for manager API using Python library ctypes and also a corresponding solverdummy

erikscheurer:
This PR creates a solverdummy in C++ for the micromanager. 

In a new example, the c++ dummy is just the python solverdummy "translated" to c++. Couple c++ to python, we need to "expose" the coupling functions to python, through a shared library. For this reason, we have a class `Microsimulation` in C++ as well as a class in Python that loads the shared library and calls all functions through `ctypes`. We have to do a type conversion before passing the data to the solve function.

At the moment, the data is passed as a python dictionary to C++, but as there is no equivalent to it for `ctypes`, the data is passed as separate arguments. This can be a problem if there is more data to pass, and a potential user would have to keep this in mind. 
Another problem is passing numpy arrays to C++, from what I could find, the best way is to pass by reference. This means we don't use `std::vector` may be preferred?

This approach is very much up to debate as in #17. 
This implementation is fairly easy for the creation of the micromanager, but may not be optimal for the user, as a python and a c++ class have to be written or modified, including possibly a lot of debugging if not already familiar with `ctypes`


erikscheurer:
At the moment, the dummy run fine with 
```bash
g++ -std=c++11 -shared -fPIC micro_dummy.cpp -o micro_dummy.so
python run_micro_manager.py
```
and 
```bash
python macro_dummy.py
```

IshaanDesai:
Thanks for starting to work on this! There are a couple of points to discuss before we finalize things here:

* Let us have a folder structure which is consistent with the [solver dummies in the precice repo](https://github.com/precice/precice/tree/develop/examples/solverdummies) where only the relevant files are kept in folders with titles of the language that they are written in, and the preCICE config file and README and kept centrally one level above. This way we will not need to add another Python macro code to be run with a C++ micro code.
* The functionality in the `micro_dummy_wrapper.py` should go in the Micro Manager source code. The idea is that the Micro Manager detects during configuration (the initialization step) whether the file it is getting is a Python script or a compiled code, and then accordingly uses the appropriate functionality. We can sketch the configuration in-person and then you can implement it here.
* This PR should eventually also add CI to run tests for combinations of the solver dummy, so python-macro - python-micro (already exists) and python-macro - c++-micro.
* Regarding the problem of passing Python dictionaries to C++, I think an easy solution does not exist. The Micro Manager can handle this, meaning as it already knows the language of the micro code, it can either pass a Python dictionary, or simply a NumPy array.
* Whether `std::vector` would be preferred or not, its not easy to say. We have to remember that the user should have as much flexibility as possible in writing their micro code. This means that we should pass something that is easy to work with.

micro_manager/micro_cpp_wrapper.py
```diff
@@ -0,0 +1,31 @@
...
+        self.MicroSimulation_lib.MicroSimulation_solve.argtypes = [ctypes.c_void_p, ctypes.POINTER(ctypes.c_double), ctypes.POINTER(ctypes.c_double), ctypes.c_double]
+        # return type is a std::pair of float and numpy array
+        self.obj = self.MicroSimulation_lib.MicroSimulation_new(sim_id)
+
+    def initialize(self):
+        self.MicroSimulation_lib.MicroSimulation_initialize(self.obj)
+
+    def solve(self, macro_data, dt:float):
+        macro_scalar_data_ptr = ctypes.c_double(macro_data["macro-scalar-data"])
+        macro_vector_data_ptr = macro_data["macro-vector-data"].ctypes.data_as(ctypes.POINTER(ctypes.c_double))
```
erikscheurer, line 18: Currently only made for the solverdummy data. We have to extract the data from the config file and adjust the data that is passed to c++ accordingly.

IshaanDesai:
@erikscheurer can you please move the development done on the branch on your fork to a branch on the main repo? Then we can close this PR and keep the branch in case we want to have a look at it in the future.

erikscheurer:
The code is now also on [a repo branch](https://github.com/precice/micro-manager/tree/c++-sharedObject)

PR #22, opened 2023-02-09
Add C++ bindings for manager API using pybind11 and also a corresponding solverdummy

erikscheurer:
This PR adds a C++ solverdummy using pybind11 as discussed in #17. 
The binding with pybind11 seems far easier to us and to users. 
A user could for example use this dummy as a wrapper for their own code or create their own interface with pybind11 by just adding a few lines at the end of their c++ class as in the cpp_dummy here. Compilation is also as easy as with the other approach from #20 

Passing dictionaries between python and c++ is as easy as using `pybind11::dict`. However, for type conversion of the passed data, the user may have to tweak the dummy. How difficult these changes would be, I can't say at the moment without an example. Casting is in any case easier than with the `ctypes` approach from #20.


IshaanDesai:
Can I already merge this? Also, please merge develop into your branch and overwrite the file `examples/macro-micro-dummy/README.md` that I have created because what you have written up is much better.

erikscheurer:
No not yet, we observed some weird behavior when increasing the dummy data in each step, that we should maybe test more.

examples/README.md
```diff
@@ -0,0 +1,42 @@
+# Solverdummies
+
+The `solverdummies` are minimal working examples for using the preCICE micro-manager with different languages. At the moment, there are examples for Python, and C++. They can be coupled with any other solver, for example the `macro-dummy.py` in this directory.
```
IshaanDesai, line 3:
```suggestion
The `solverdummies` are minimal working examples for using the preCICE Micro Manager with different languages. At the moment, there are examples for Python, and C++. They can be coupled with any other solver, for example the `macro-dummy.py` in this directory.
```

examples/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Python
+To run the Python solverdummies, run the following commands in **two different terminals** the `examples` directory:
```
IshaanDesai, line 6:
```suggestion
To run the Python solverdummies, run the following commands in the `examples/` directory in **two different terminals**:
```

examples/README.md
```diff
@@ -0,0 +1,42 @@
...
+<details>
+<summary>Explanation</summary>
+
+The command above compiles the C++ solverdummy and creates a shared library that can be imported from python using `pybind11`. 
+- The `$(python3 -m pybind11 --includes)` part is necessary to include the correct header files for `pybind11`. 
+- The `$(python3-config --extension-suffix)` part is necessary to create the correct file extension for the shared library. For more information, see the [pybind11 documentation](https://pybind11.readthedocs.io/en/stable/compiling.html#building-manually).
+
+</details>
+
+Then, run the following commands in **two different terminals** the `examples` directory:
```
IshaanDesai, line 34:
```suggestion
Then, run the following commands in the `examples/` directory, in **two different terminals**:
```

examples/cpp-dummy/micro_cpp_dummy.cpp
```diff
@@ -0,0 +1,111 @@
...
+    void reload_checkpoint();
+    int get_dims();
+
+private:
+    int _sim_id;
+    int _dims;
+    double _micro_scalar_data;
+    std::vector<double> _micro_vector_data;
+    double _checkpoint;
+};
```
IshaanDesai, line 29: Does it make more C++ sense to move this to a `micro_cpp_dummy.hpp` file? If yes, lets do it here too, to be consistent with what a developer in C++ would do.

examples/cpp-dummy/micro_cpp_dummy.cpp
```diff
@@ -0,0 +1,111 @@
...
+    _micro_vector_data = std::vector<double>(macro_vector_data.data(), macro_vector_data.data() + macro_vector_data.size()); // convert numpy array to std::vector. directly casting to std::vector does not work?
+
+    // micro_scalar_data and micro_vector_data are writedata+1
+    _micro_scalar_data = macro_scalar_data + 1.;
+    for (uint i = 0; i < _micro_vector_data.size(); i++)
+    {
+        _micro_vector_data[i] += 1.;
+    }
+
+    // create python dict for micro_write_data
```
IshaanDesai, line 69:
```suggestion
    // Convert data to a py::dict again to send it back to the Micro Manager
```

examples/cpp-dummy/micro_cpp_dummy.cpp
```diff
@@ -0,0 +1,111 @@
...
+void MicroSimulation::reload_checkpoint()
+{
+    std::cout << "Reverting to old state of micro problem (" << _sim_id << ")\n";
+    _micro_scalar_data = _checkpoint;
+}
+
+int MicroSimulation::get_dims()
+{
+    return _dims;
+}
```
IshaanDesai, line 95:
```suggestion
```

examples/cpp-dummy/micro_cpp_dummy.cpp
```diff
@@ -0,0 +1,111 @@
...
+class MicroSimulation
+{
+public:
+    MicroSimulation(int sim_id);
+    void initialize();
+    // solve takes python dict for macro_write data, dt, and returns python dict for macro_read data
+    py::dict solve(py::dict macro_write_data, double dt);
+    void save_checkpoint();
+    void reload_checkpoint();
+    int get_dims();
```
IshaanDesai, line 21:
```suggestion
```

examples/cpp-dummy/micro_cpp_dummy.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+private:
+    int _sim_id;
+    int _dims;
```
IshaanDesai, line 25:
```suggestion
```

IshaanDesai (changes requested):
Good job! :clap: just some minor points to be done, mostly cleanup. I guess it possible to also have only one `micro-manager-config.json` file, which can then be in the `examples/` directory. Can you check if this works, and if it does, move the file?

PR #26, opened 2023-03-08
Do some syntax changes

erikscheurer:
This PR cleans up small things in the code of the micromanager. The functionality is not supposed to be changed here.

- replace manual `counter` wit `enumerate()`
- use numpy vector access to matrices
- use `str().replace()`
- use `with` syntax for reading files


IshaanDesai (approved):
Thanks for the updates!

PR #30, opened 2023-04-14
Add pickling support to C++ solver dummy

IshaanDesai:
This PR adds functionality to allow for pickling and unpickling of the C++ micro simulation code in the C++ solver dummy. This is based on the [pickling support documentation of pybind11](https://pybind11.readthedocs.io/en/latest/advanced/classes.html#pickling-support).

### DIFF
--- a/examples/cpp-dummy/micro_cpp_dummy.cpp
+++ b/examples/cpp-dummy/micro_cpp_dummy.cpp
@@ -69,6 +69,19 @@ void MicroSimulation::reload_checkpoint()
     _micro_scalar_data = _checkpoint;
 }
 
+// This function is not relevant for the Micro Manager but is important for serialization of the micro simulation object
+void MicroSimulation::setState(double micro_scalar_data, double checkpoint)
+{
+    _micro_scalar_data = micro_scalar_data;
+    _checkpoint = checkpoint;
+}
+
+std::tuple<double, double> MicroSimulation::getState()
+{
+    return {_sim_id, _micro_scalar_data, _checkpoint};
+}
+
+
 PYBIND11_MODULE(micro_dummy, m) {
     // optional docstring
     m.doc() = "pybind11 micro dummy plugin";
@@ -78,5 +91,22 @@ PYBIND11_MODULE(micro_dummy, m) {
         .def("initialize", &MicroSimulation::initialize)
         .def("solve", &MicroSimulation::solve)
         .def("save_checkpoint", &MicroSimulation::save_checkpoint)
-        .def("reload_checkpoint", &MicroSimulation::reload_checkpoint);
+        .def("reload_checkpoint", &MicroSimulation::reload_checkpoint)
+        .def(py::pickle(
+            [](const MicroSimulation &ms) { // __getstate__
+                /* Return a tuple that fully encodes the state of the object */
+                return py::make_tuple(ms.getState());
+            },
+            [](py::tuple t) { // __setstate__
+                if (t.size() != 3)
+                    throw std::runtime_error("Invalid state!");
+                
+                /* Create a new C++ instance */
+                MicroSimulation ms(t[0]);
+
+                ms.setState(t[1].cast<double>(), t[2].cast<double>());
+
+                return ms;
+            }
+        ));
 }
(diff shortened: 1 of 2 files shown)


examples/cpp-dummy/micro_cpp_dummy.cpp
```diff
@@ -69,6 +69,21 @@ void MicroSimulation::reload_checkpoint()
     _micro_scalar_data = _checkpoint;
 }
 
+// This function needs to set the complete state of a micro simulation
+void MicroSimulation::setState(double micro_scalar_data, double checkpoint)
```
erikscheurer, line 73: Do we here also need `sim_id` and `micro_vector_data?`

erikscheurer, line 73: Same as above, is micro_vector_data not needed to fully define the state?

erikscheurer: This comment was supposed to be at [line 83](https://github.com/precice/micro-manager/blob/841c32510cee3357a2122d7ae7460ec2638b001f/examples/cpp-dummy/micro_cpp_dummy.cpp#L83)

erikscheurer:
Some questions from my side, otherwise it looks good.
I think it would be very useful for this PR to be also merged into #29, because if we have a get and set state method, implementing the `__deepcopy__` function would be as easy as saying, create a new instance with `self._sim_id`, `get_state` from self and `set_state` of the new instance. This could then be part of the c++ dummy and used without modification by the user or could even be ignored as written in the [pybind deepcopy documentation](https://pybind11.readthedocs.io/en/latest/advanced/classes.html#deepcopy-support).
If we write in the documentation that the simulation state should be fully contained in the set_state method, it is more clear that not every variable of the object has to be copied as I did [in the deepcopy implementation](https://github.com/precice/micro-manager/blob/e76ff83da955eaeb22a239de6b81acab37d46049/examples/cpp-dummy/micro_cpp_dummy.cpp#L73-L80) with the `checkpoint` as well, but only the necessary ones.

erikscheurer:
I get an invalid state error running:
```python
import pickle
import micro_dummy
sim = micro_dummy.MicroSimulation(2)
pickle.loads(pickle.dumps(sim))
```
I think this is because `py::make_tuple` is called on a `std::tuple` returning a tuple of a tuple ie length 1.
Moving the make_tuple call to the get_state function seemed to solve the issue for me, but I don't know if this is the intended definition of the getState function.

examples/cpp-dummy/micro_cpp_dummy.cpp
```diff
@@ -69,6 +69,21 @@ void MicroSimulation::reload_checkpoint()
...
+{
+    _micro_scalar_data = micro_scalar_data;
+    _checkpoint = checkpoint;
+}
+
+
+// This function needs to return variables which can fully define the state of a micro simulation
+std::tuple<double, double, double> MicroSimulation::getState() const
```
erikscheurer, line 81:
```suggestion
py::tuple MicroSimulation::getState() const
{
    return py::make_tuple(_sim_id, _micro_scalar_data, _checkpoint);
}
```

examples/cpp-dummy/micro_cpp_dummy.cpp
IshaanDesai: This is just to show an example to show the user on how to use the functions. We don't actually need to set and get the full state of a dummy problem. Hence I have purposely kept things simple here.

erikscheurer (approved):

examples/cpp-dummy/micro_cpp_dummy.cpp
```diff
@@ -78,5 +91,22 @@ PYBIND11_MODULE(micro_dummy, m) {
         .def("initialize", &MicroSimulation::initialize)
         .def("solve", &MicroSimulation::solve)
         .def("save_checkpoint", &MicroSimulation::save_checkpoint)
-        .def("reload_checkpoint", &MicroSimulation::reload_checkpoint);
+        .def("reload_checkpoint", &MicroSimulation::reload_checkpoint)
+        .def(py::pickle(
+            [](const MicroSimulation &ms) { // __getstate__
+                /* Return a tuple that fully encodes the state of the object */
+                return py::make_tuple(ms.getState());
```
erikscheurer, line 98:
```suggestion
                return ms.getState();
```

PR #32, opened 2023-04-14
Add global variant to adaptivity

IshaanDesai:
This PR adds a global variant to the already existing local adaptivity. Local adaptivity means that micro simulations within a rank are compared for similarity, whereas global adaptivity means that micro simulations in the entire domain are compared for similarity.


IshaanDesai:
Closed in favor of https://github.com/precice/micro-manager/pull/42

PR #34, opened 2023-04-14
Resolve bug in copy operation when inactive simulation is activated

IshaanDesai:
This PR sets local and global ID of copied active simulation to the original IDs of the previously inactive simulation. If we do not do this, the newly activated micro simulation will have the same local and global ID as the active simulation which it was copied from.

First stopped via https://github.com/precice/micro-manager/pull/29#issuecomment-1508148493

### DIFF
--- a/micro_manager/adaptivity.py
+++ b/micro_manager/adaptivity.py
@@ -161,11 +161,17 @@ class AdaptiveController:
                 if self._check_for_activation(i, similarity_dists, _micro_sim_states):
                     associated_active_id = micro_sims[i].get_associated_active_id()
 
+                    # Get local and global ID of inactive simulation, to set it to the copied simulation later
+                    local_id = micro_sims[i].get_local_id()
+                    global_id = micro_sims[i].get_global_id()
+
                     # Effectively kill the micro sim object associated to the inactive ID
                     micro_sims[i] = None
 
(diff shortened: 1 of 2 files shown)


erikscheurer:
The microproblem extends the `base_micro_simulation` and initializes it with `local_id` [here](https://github.com/precice/micro-manager/blob/f016b42c0b6e3f1aa39f3c4e2180228fc282b301/micro_manager/micro_manager.py#L38).
Why do we need to pass the `local_id` to the base simulation, is this something that will be used for actual microsimulations? or is it just for debugging as in the [`micro-dummy`](https://github.com/precice/micro-manager/blob/f016b42c0b6e3f1aa39f3c4e2180228fc282b301/examples/python-dummy/micro_dummy.py#L13). Here the `sim_id` is only used for printing but is also deep-copied over from the active simulation to the inactive one. If this sim_id is necessary, we also need to change this id in the base_micro_simulation.

IshaanDesai:
> Why do we need to pass the local_id to the base simulation, is this something that will be used for actual micro simulations?

This is a design decision that is yet to be finalized. Initially the idea of passing the local ID to the micro simulations was to get them to be unique, in the event that the user might want to output something from them for debugging purposes. But this is not a strong argument and I could imagine that in the long term, passing the ID is not done.

PR #35, opened 2023-04-17
Micro Manager no longer passes local ID to micro simulation during object creation

IshaanDesai:
Until now the Micro Manager used to pass the local ID of a micro simulation object to the object constructor at the time of its creation. This is problematic for parallel simulations and also for adaptivity, and hence is being removed. A possible drawback of this change is that now the micro simulation objects themself are not uniquely identifiable. This could be a problem if the output functionality in the micro simulation is called. Further discussion is necessary on whether it is relevant or not to have some uniqueness definition for the micro simulation objects.


erikscheurer (approved):
Looks good. One suggestion, we could maybe offer a optional `set_sim_id` function for debbuging? Meaning that the micromanager checks if the given MicroSimulation has a function `set_sim_id` and if it does it passes the local id. We still run into the issues with copying and what you described but I think if we can fix these I think it can definitely be helpful for debbuging.
On the other hand it may make the documentation more cluttered.

HelenaKschidock:
Ok, I will adapt the Dumux sim accordingly. Agree with Eric that readding some sort of identifier in the future might be nice for users, if you find an elegant solution to do so

PR #37, opened 2023-05-04
Add unit tests for micro_manager file

### LINKED ISSUE #19: Add unit tests for all user side functions of the Micro Manager
All the functions in the file `micro_manager/micro_manager.py` needs to have unit tests for testing their behavior in several situations.

erikscheurer:
closes #19 
This PR adds unittests for `MicroManager.__init__`, `initialize`, `read_data_from_precice`, `write_data_from_precice`, `solver_micro_simulations` and `Config.read_json`


.github/workflows/check-pep8.yml
```diff
@@ -2,7 +2,8 @@ name: autopep8
 on:
   push:
     branches:
-      - "*"
+      - main
+      - develop
```
IshaanDesai, line 6: Is this change actually relevant to this PR? And why not check the formatting for other branches?

.github/workflows/run-adaptivity-test.yml
```diff
@@ -2,7 +2,8 @@ name: Run tests for adaptivity
 on:
   push:
     branches:
-      - "*"
+      - main
+      - develop
```
IshaanDesai, line 6: Would suggest pushing some changes via other PRs (or direct commits to develop, if the change is straightforward), rather than part of a PR where this is not directly relevant.

tests/unit/precice.py
```diff
@@ -0,0 +1,80 @@
+# This file imitates the python bindings for precice and is used _only_ to test the micro-manager
```
IshaanDesai, line 1:
```suggestion
# This file mocks  pyprecice, the Python bindings for preCICE and is used _only_ for unit testing the Micro Manager.
```

tests/unit/test_functioncalls.py
```diff
@@ -0,0 +1,73 @@
...
+        assert macro_data["macro-vector-data"].tolist() == [0, 1, 2]
+        return {"micro-scalar-data": macro_data["macro-scalar-data"] + 1,
+                "micro-vector-data": macro_data["macro-vector-data"] + 1}
+
+
+class TestFunctioncalls(TestCase):
+
+    def test_micromanager_initialize(self):
+        manager = micro_manager.MicroManager('test_unit.json')
+        self.assertTrue(True)
```
IshaanDesai, line 24: What are we testing here? Is it the constructor of the MicroManager class? I would add descriptions to all the tests to shortly explain what is being tested.

tests/unit/test_functioncalls.py
```diff
@@ -0,0 +1,73 @@
...
+        read_data = manager.read_data_from_precice()
+        for i in range(3):
+            self.assertEqual(read_data[i]["macro-scalar-data"], 1)
+            self.assertListEqual(read_data[i]["macro-vector-data"].tolist(), [0, 1, 2])
+
+    def test_solve_mico_sims(self):
+        manager = micro_manager.MicroManager('test_unit.json')
+        manager.initialize()
+        micro_sims_output = manager.solve_micro_simulations(
+            [{"macro-scalar-data": 1, "macro-vector-data": np.array([0, 1, 2]), "micro_sim_time": 0}] * 4, np.array([True, True, True, True]))
```
IshaanDesai, line 54: Similar to the comment above, lets create a `fake_micro_sims_input` and a `fake_micro_sims_output`.

IshaanDesai (changes requested):
Looks quite good already! Lets implement the changes I have suggested before finalizing this.

.github/workflows/check-pep8.yml
erikscheurer: Sorry, forgot to remove this change. If we run tests on all branches when pushing, the tests are executed twice. One time because we push and the other time because we probably created a pull request.
It is not necessary to make this change, but I think it may makes sense as it reduces computation time. The [main repo](https://github.com/precice/precice/blob/8da078b35efd7fe91838c4996bbb5e7658f3e7bb/.github/workflows/build-and-test.yml#L8-L23) also does it this way.

tests/unit/test_functioncalls.py
erikscheurer: Yes it tests the constructor, do you think it would also make sense to check values here?

.github/workflows/check-pep8.yml
IshaanDesai: Makes sense, can you please open a separate PR with this change?

tests/unit/test_functioncalls.py
IshaanDesai: Yes lets check some values which are set inside the constructor rather than only checking if the constructor runs through. For example, any of the internal variables that are set based on what is provided in the configuration.

tests/unit/precice.py
```diff
@@ -0,0 +1,80 @@
...
+    def mark_action_fulfilled(self, action):
+        pass
+
+    def initialize_data(self):
+        pass
+
+    def write_block_scalar_data(self, data_id, vertex_ids, data):
+        if data_id == 1:  # micro-scalar-data not micro_sim_time
+            self.read_write_scalar_buffer = data
+        print("write_block_scalar_data", data)
```
IshaanDesai, line 48:
```suggestion
```

tests/unit/precice.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def write_block_vector_data(self, data_id, vertex_ids, data):
+        self.read_write_vector_buffer = data
+        print("write_block_vector_data", data)
```
IshaanDesai, line 52:
```suggestion
```

tests/unit/test_functioncalls.py
```diff
@@ -0,0 +1,79 @@
...
+        self.assertListEqual(manager._macro_bounds, self.macro_bounds)
+        self.assertListEqual(manager._mesh_vertex_ids.tolist(), [0, 1, 2, 3])
+        self.assertEqual(len(manager._micro_sims), 4)
+        self.assertEqual(manager._micro_sims[0].very_important_value, 0)  # test inheritance
+        self.assertDictEqual(manager._read_data_names, self.fake_read_data_names)
+        self.assertDictEqual(dict(self.fake_write_data_names, **{'micro_sim_time': False}), manager._write_data_names)
+
+    def test_read_write_data_from_precice(self):
+        manager = micro_manager.MicroManager('test_unit.json')
+        manager.initialize()
```
IshaanDesai, line 50: In unit testing, only the call to the function that is being tested should be done. If internal variables need to be set before, they are set manually. Lets remove the call to `initialize()` here, to make the test simpler.

tests/unit/test_functioncalls.py
```diff
@@ -0,0 +1,79 @@
...
+        manager.write_data_to_precice(self.fake_write_data)
+        read_data = manager.read_data_from_precice()
+        for data, fake_data in zip(read_data, self.fake_write_data):
+            self.assertEqual(data["macro-scalar-data"], 1)
+            self.assertListEqual(data["macro-vector-data"].tolist(),
+                                 fake_data["micro-vector-data"].tolist())
+
+    def test_solve_mico_sims(self):
+        manager = micro_manager.MicroManager('test_unit.json')
+        manager.initialize()
```
IshaanDesai, line 60: Same as above.

tests/unit/test_functioncalls.py
```diff
@@ -0,0 +1,79 @@
...
+        micro_sims_output = manager.solve_micro_simulations(self.fake_read_data, np.array([True, True, True, True]))
+        for data, fake_data in zip(micro_sims_output, self.fake_write_data):
+            self.assertEqual(data["micro-scalar-data"], 2)
+            self.assertListEqual(data["micro-vector-data"].tolist(),
+                                 (fake_data["micro-vector-data"] + 1).tolist())
+
+    def test_config(self):
```
IshaanDesai, line 67: Lets add a variant with adaptivity here, so `test_config_without_adaptivity` and `test_config_with_adaptivitiy`.

IshaanDesai (changes requested):
Looks good! Final changes needed now.

tests/unit/test_functioncalls.py
erikscheurer: We seem to be needing `initialize` for reading and writing see [here](https://github.com/precice/micro-manager/actions/runs/5044114144/jobs/9046745186?pr=37#step:5:25)

tests/unit/test_functioncalls.py
erikscheurer: We seem to be needing `initialize` for solve_micro_sims see [here](https://github.com/precice/micro-manager/actions/runs/5044114144/jobs/9046745186?pr=37#step:5:25). Should we just set the local number of micro_sims by hand?

tests/unit/test_functioncalls.py
IshaanDesai: Good point, lets have only a case with adaptivity here.

PR #40, opened 2023-05-15
Add user option for similarity measures

### LINKED ISSUE #24: Allow user to provide comparison function for similarity calculation in adaptivity
Currently the adaptivity calculates similarity between two simulations based on an absolute difference:

https://github.com/precice/micro-manager/blob/cec8910d8cefbdb32465b520821e9a1069219414/micro_manager/adaptivity.py#L59-L62

Instead of doing an absolute difference, the user can provide a function of their choice to use in the similarity computation. Alternatively the manager can provide various options for the user to pick from.

erikscheurer:
Closes #24 

This PR adds options for different similarity measures. Currently L1 and L2 are supported.


erikscheurer:
When implementing a relative L1 norm for example, by which quantity do we divide? For example, we have the two simulation with states `1` and `2`, then `data=[1,2]`. (So still a single scalar value to compare simulations)
If we then calculate the distances between the two we would have `diff =[[0 , 1-2],[2-1, 0]]` if we do not normalize. How would this look normalized? Normalization by the maximum of the two? If we do a pointwise division with a vector, we would not get a symmetric matrix which does not make much sense right?

erikscheurer:
I see two options now:
1. Divide by the mean, this would mean, that different units for example would be cancelled out while preserving the symmetry in the distances. In our example this would be `diff=[[0 (1-2)/1.5], [(2-1)/1.5, 0]]`
2. Divide by the metric of the "fist" simulation. Divide the matrix row-wise with the vector. `diff=[[0, (1-2)/1], [(1-2)/2, 0]]`. This would make sense to do because the check if the simulation should be turned active or inactive is done using the row index. To check if the simulation number 0 with data `1` should be turned off, we check if `diff[0,:]<tol`, therefore we should divide row 0 by the data of simulation 0

IshaanDesai:
Lets go for the second option, so dividing my the data value of one of the simulations. Which one to pick is our choice, we should only be consistent. As for vectorial quantities, lets say we have a quantity $v$ which we want to use for the comparison. For two micro simulations `1` and `2`, this would mean that we have two vectors to compare, $$ v^{1}_{x}, v^{1}_{y}, v^{1}_{z} $$ and $$ v^{2}_{x}, v^{2}_{y}, v^{2}_{z} $$. Let us do a component-wise relative computation then, so we do $\frac{abs(v^{1}_{x} - v^{2}_{x})}{v^{1}_{x}}$ for all components, and then sum up the relative errors.

erikscheurer:
This should work now as we discussed, do we want unit tests for this?

micro_manager/adaptivity/adaptivity.py
```diff
@@ -107,3 +95,36 @@ def _check_for_deactivation(
                 if similarity_dists[active_id, active_id_2] < self._coarse_tol:
                     return True
         return False
+
+    def _get_similarity_measure(self, similarity_measure):
```
IshaanDesai, line 99: I see that this functions is called once with a `similarity_measure` string and then with `data`. Just for my understanding, how does this work exactly?

micro_manager/adaptivity/adaptivity.py
```diff
@@ -107,3 +95,36 @@ def _check_for_deactivation(
...
+            return lambda data: np.linalg.norm(data[np.newaxis, :] - data[:, np.newaxis], ord=1, axis=-1)
+        elif similarity_measure == 'L2':
+            return lambda data: np.linalg.norm(data[np.newaxis, :] - data[:, np.newaxis], ord=2, axis=-1)
+        elif similarity_measure == 'L1rel':
+            def l1rel(data):
+                pointwise_diff = data[np.newaxis, :] - data[:, np.newaxis]
+                # divide by data to get relative difference
+                # transpose to divide row i by data[i].
+                relative = np.nan_to_num((pointwise_diff.transpose(1, 0, 2) / data).transpose(1, 0, 2))
+                return np.linalg.norm(relative, ord=1, axis=-1)
```
IshaanDesai, line 118: Lets define this as a standalone function.

micro_manager/adaptivity/adaptivity.py
```diff
@@ -107,3 +95,36 @@ def _check_for_deactivation(
...
+            return l1rel
+        elif similarity_measure == 'L2rel':
+            def l2rel(data):
+                pointwise_diff = data[np.newaxis, :] - data[:, np.newaxis]
+                # divide by data to get relative difference
+                # transpose to divide row i by data[i]
+                relative = np.nan_to_num((pointwise_diff.transpose(1, 0, 2) / data).transpose(1, 0, 2))
+                return np.linalg.norm(relative, ord=2, axis=-1)
+            return l2rel
```
IshaanDesai, line 127: Same as above.

IshaanDesai (commented):
Thanks for this! These additions will expand the adaptivity functionality significantly. Minor changes from my side.

micro_manager/adaptivity/adaptivity.py
erikscheurer: `_get_similarity_measure` returns a function which is saved as `self._similarity_measure`. This function is then called with `data` not `_get_similarity_measure`.
Or did you mean the similarity measure itself?

erikscheurer:
In the future we may want to switch to a different method of selecting functions. As suggested in #24, the user could provide their own similarity functions which cannot easily be realized with the string-selection.

erikscheurer:
Do we want to merge this or wait for the documentation and the unit tests? Some documentation would be necessary for the similarity functions.

IshaanDesai:
> Do we want to merge this or wait for the documentation and the unit tests? Some documentation would be necessary for the similarity functions.

Let me finish the documentation PR, and then we can add documentation here, and merge it.

CHANGELOG.md
```diff
@@ -2,14 +2,15 @@
 
 ## latest
 
+- Add support for L2 and relative similarity functions https://github.com/precice/micro-manager/pull/40
```
IshaanDesai, line 5:
```suggestion
- Add norm-based (L1 and L2) support for functions in similarity distance calculation with absolute and relative variants https://github.com/precice/micro-manager/pull/40
```

docs/usage-configuration.md
```diff
@@ -43,6 +43,7 @@ Parameter | Description
 Parameter | Description
 --- | ---
 `macro_domain_bounds`| Minimum and maximum bounds of the macro-domain, having the format `[xmin, xmax, ymin, ymax, zmin, zmax]` in 3D and `[xmin, xmax, ymin, ymax]` in 2D.
+*optional:* `adaptivity_similarity_measure`| Similarity measure to be used for adaptivity. Can be either `L1`, `L2`, `L1rel` or `L2rel`. By default, `L1` is used. The `rel` variants are the relative versions of the `L1` and `L2` norms, dividing the difference by the maximum value of the data.
```
IshaanDesai, line 46:
```suggestion
*optional:* `adaptivity_similarity_measure`| Similarity measure to be used for adaptivity. Can be either `L1`, `L2`, `L1rel` or `L2rel`. By default, `L1` is used. The `rel` variants calculate the respective relative norms.
```

micro_manager/adaptivity/adaptivity.py
```diff
@@ -107,3 +95,49 @@ def _check_for_deactivation(
...
+        elif similarity_measure == 'L1rel':
+            return self._l1rel
+        elif similarity_measure == 'L2rel':
+            return self._l2rel
+        else:
+            raise ValueError(
+                'Similarity measure not supported. Currently supported similarity measures are "L1", "L2", "L1rel", "L2rel".')
+
+    def _l1(self, data: np.ndarray) -> np.ndarray:
+        return np.linalg.norm(data[np.newaxis, :] - data[:, np.newaxis], ord=1, axis=-1)
```
IshaanDesai, line 126: Can you please add a docstring block to all similarity measure functions, similar to the ones that we have for all other functions?

IshaanDesai (approved):
Looks good :+1: only a few documentation related comments from my side. After adding the documentation we can merge.

docs/usage-configuration.md
erikscheurer: Do we not keep the explanation of the "relative-ness" in here?

IshaanDesai: > Do we not keep the explanation of the "relative-ness" in here?

That much detail is not necessary here. If someone really wants to know how we compute the relative norm, they are free to look into the code. What we do is per se not special.

micro_manager/adaptivity/adaptivity.py
```diff
@@ -36,25 +39,10 @@ def get_similarity_dists(self, dt: float, similarity_dists: np.ndarray, data: np
...
-                    if dim:
-                        for d in range(dim):
-                            data_diff += abs(data[counter_1, d] - data[counter_2, d])
-                    else:
-                        data_diff = abs(data[counter_1] - data[counter_2])
-
-                    _similarity_dists[id_1, id_2] += dt * data_diff
-                else:
-                    _similarity_dists[id_1, id_2] = 0
+            data = np.expand_dims(data, axis=1)
```
IshaanDesai, line 42: Please write a comment here on what you are actually doing. Bigger description is also okay.

micro_manager/adaptivity/adaptivity.py
```diff
@@ -107,3 +95,103 @@ def _check_for_deactivation(
...
+        ----------
+        data : numpy array
+            Data to be used in similarity distance calculation
+
+        Returns
+        -------
+        similarity_dists : numpy array
+            Updated 2D array having similarity distances between each micro simulation pair
+        """
+        return np.linalg.norm(data[np.newaxis, :] - data[:, np.newaxis], ord=1, axis=-1)
```
IshaanDesai, line 139: Please write a comment about what the `ord`, `axis` and `newaxis` are doing. I know that this seems like an overkill, but it really helps a new user in understanding what is going on.

IshaanDesai (approved):

PR #43, opened 2023-05-19
Change branches to run tests on push

erikscheurer:
This PR changes on which branches the tests are being run. When its `"*"` the tests are run twice for each PR. Therefore only testing on each push for `develop` and `main` is preferred.
The changed is discussed in #37

### DIFF
--- a/.github/workflows/check-pep8.yml
+++ b/.github/workflows/check-pep8.yml
@@ -2,7 +2,8 @@ name: autopep8
 on:
   push:
     branches:
-      - "*"
+      - main
+      - develop
   pull_request:
     branches:
       - "*"
--- a/.github/workflows/run-adaptivity-test.yml
+++ b/.github/workflows/run-adaptivity-test.yml
@@ -2,8 +2,8 @@ name: Run tests for adaptivity
 on:
   push:
     branches:
-      - "main"
-      - "develop"
+      - main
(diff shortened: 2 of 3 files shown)


.github/workflows/check-pep8.yml
```diff
@@ -2,7 +2,8 @@ name: autopep8
 on:
   push:
     branches:
-      - "*"
+      - main
```
IshaanDesai, line 5: For the other two files you have the branch names in quotes, but here not. Do we need the double quotes? If not, lets keep things clean :)

IshaanDesai (commented):
Makes sense!

.github/workflows/check-pep8.yml
erikscheurer: > The characters *, [, and ! are special characters in YAML. If you start a pattern with *, [, or !, you must enclose the pattern in quotes.

From the [github actions documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet). So sadly we need the quotes.

IshaanDesai: Okay, then lets also have quotes in the check-pep8 workflow file.

IshaanDesai (approved):

PR #51, opened 2023-06-18
Update preCICE API calls for `precice:develop`

erikscheurer:
This PR updates the micro-manager preCICE version to the current API from the [python-bindings](https://github.com/precice/python-bindings/blob/develop/cyprecice/cyprecice.pyx).

One Problem I am currently facing is requiring initial data. The Micro Manager uses `self._participant.get_mesh_vertices_and_ids` to get the number of local micro sims. However, now that `initialize_data` and `initialize` is just `initialize` we can't do:
1. `initialize`
2. `get_mesh_vertices_and_ids`
3. initialize all micro_sims and get initial data
4. `initialize_data`

as we did before.
Related to that problem are probably the changes in the precice-config where some precice-error occured with
```xml
       <exchange data="micro-scalar-data" mesh="macro-mesh" from="Micro-Manager" to="Macro-dummy" initialize="yes"/>
```


IshaanDesai:
Thanks for already opening the v3 update PR!

As far as I know, the function `get_mesh_vertices_and_ids` should work before initializing preCICE. So the new order of doing things would be:

1. Call `get_mesh_vertices_and_ids` to get the macro vertices
2. Initialize all micro simulations and get the initial states if the user has provided them.
3. Check for each data if preCICE requires initial data, using the API function `requires_initial_data`
4. If initial data is required, explicitly write the data using `write_data`

> Also the tests currently fail due to an error in installing the python bindings on the develop container

Weird, I will have a look here.

IshaanDesai:
As stated in https://github.com/precice/micro-manager/issues/54, the Micro Manager can no longer write initial data to preCICE. The `initialize()` function requirement in the library form of the micro simulations is henceforth dropped.

IshaanDesai (approved):

PR #83, opened 2024-03-12
Remove unused import of micro simulations

tjwsch:
This removes an unnecessary import of the `MicroSimulation` class that was not used. In a personal meeting, we also discussed if one of the logger level sets could be removed but on further investigation, they seem to serve slightly different purposes.

### DIFF
--- a/micro_manager/micro_manager.py
+++ b/micro_manager/micro_manager.py
@@ -71,7 +71,6 @@ class MicroManager:
             self._size)
 
         micro_file_name = self._config.get_micro_file_name()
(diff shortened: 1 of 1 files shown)


IshaanDesai (commented):
Can you also please remove the other lines that we discussed? Good to do more than one cleanups in one PR.

tjwsch:
After further investigation on the other lines, I found that setting the logger level in the root logger and a handler serves different purposes, thus both `setLevel` calls are necessary, as I wrote in my first comment. That was the only other cleanup I remember us discussing.

IshaanDesai (approved):

PR #84, opened 2024-03-19
Add pyproject.toml

IshaanDesai:


pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+]
+
+[project.urls]
+Homepage = "https://precice.org"
+Documentation = "https://precice.org/tooling-micro-manager-overview.html"
+Repository = "https://github.com/precice/micro-manager"
+"Bug Tracker" = "https://github.com/precice/micro-manager/issues"
+
+[project.scripts]
+micro_manager_precice = "micro_manager.micro_manager:main"
```
fsimonis, line 37: Is the name change of the executable intentional?

pyproject.toml
```diff
@@ -0,0 +1,43 @@
+[build-system]
+requires = ["setuptools>=41", "wheel", "setuptools-git-versioning"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name="micro-manager-precice"
+dynamic = [ "version" ]
+dependencies = [
+    "pyprecice>=3.0.0.0", "numpy", "mpi4py", "scikit-learn"
```
fsimonis, line 9: You could request a compatible release instead.

```suggestion
    "pyprecice~=3.0", "numpy", "mpi4py", "scikit-learn"
```

fsimonis (approved):

pyproject.toml
IshaanDesai: Yes, I want to make it consistent with the package name. In any case, this name should not be directly used anywhere as per my understanding.

pyproject.toml
IshaanDesai: I could not find a reference to this syntax. How does it change what was happening before?

fsimonis: https://peps.python.org/pep-0440/#compatible-release

pyproject.toml
fsimonis: A tutorial uses it, but this can be fixes easily.

IshaanDesai: > A tutorial uses it, but this can be fixed easily.

Yes, right. Actually using the executable is the best way to use the tool, so I now made it the default way to use the Micro Manager. I will open another pull request in the tutorials repo to change the name.

Issue #74, opened 2024-01-29
Handle crashing of micro simulations in a proper way


IshaanDesai:
Currently when the Micro Manager is controlling and running micro simulations, if one simulation crashes or has an improper exit, the Micro Manager run just hangs. It is nearly impossible to know which micro simulation crashed, and why the overall execution has hanged.

The Micro Manager should be able to handle a simulation crash by ...
- ... continuing to run the rest of the micro simulations.
- ... logging the simulation crash in the log file.
- ... providing the macro location at which the simulation crashed.
- ... creating a new log file and parsing the error output from the crashed simulation to it.

It is not clear if all these things could be done, so initially some investigations are necessary.

uekerman:
Ideally, the complete simulation could continue to run, by replacing the result of the failing micro simulation by a similar simulation.
Technically, this should be handled by catching exceptions.

uekerman:
> Especially in the case of adaptivity, as I understand it, simulations that are similar to each other rely on one to run, and if it hangs or crashes, these inactive simulations would not help resolve the crash/hang.

It could be the next similar one.

> Should this crashed simulation rely on data from another simulation from the point of failure until the end of the simulation, or should it attempt to restart somehow?

No restart. What we had in mind were cases failing in a deterministic way. So, restart should fail again.

> Thus, the simulation could also run into cases when it might be best to abort the entire solving process if continuing the complete simulation is unreasonable due to a lack of a similar simulation to replace the crashed one with.

Agree

>  is there a way to let the other participant know that the simulation has been stopped early and finalize the incomplete simulation properly?

Currently not, but under discussion in https://github.com/precice/precice/issues/1118

IshaanDesai:
Resolved via https://github.com/precice/micro-manager/pull/85

PR #85, opened 2024-03-19
Handle crashing micro simulations

tjwsch:
This is an attempt to solve issue #74.
This cannot be a permanent solution to the handling of crashing simulation. If a simulation crashes during the first time step, it is replaced with data from another random simulation. In the long run, I don't see any other solution than an interpolation (in cases where only one simulation is given and it cannot be replaced by one of different complexity).
Moreover, it has not been tested with real simulations, in parallel, or using spack.

### DIFF
--- a/micro_manager/interpolation.py
+++ b/micro_manager/interpolation.py
@@ -33,7 +33,7 @@ class Interpolation:
         assert len(all_local_coords) > 0, "No local coordinates provided."
         if len(all_local_coords) < k:
             self._logger.info(
-                "Number of neighbors {} is larger than the number of neighbors {}. Setting k to {}.".format(
+                "Number of desired neighbors k = {} is larger than the number of available neighbors {}. Setting k = {}.".format(
                     k, len(all_local_coords), len(all_local_coords)
                 )
             )
--- a/micro_manager/interpolation.py
+++ b/micro_manager/interpolation.py
@@ -9,75 +9,89 @@ class Interpolation:
 
     def get_nearest_neighbor_indices_local(
         self,
-        all_local_coords,
-        inter_point,
+        all_local_coords: np.ndarray,
+        inter_point: np.ndarray,
         k: int,
-        inter_point_is_neighbor: bool = False,
     ) -> np.ndarray:
         """
         Get the indices of the k nearest neighbors of a point in a list of coordinates.
-        Note: It can be chosen whether the point itself is considered as a neighbor or not.
-        Args:
-            all_local_coords: list
-                List of coordinates of all points.
-            inter_point:
-                Coordinates of the point for which the neighbors are to be found.
-            k: int
-            inter_point_is_neighbor: bool, optional
-                Decide whether the interpolation point is considered as its own neighbor.
-                Defaults to False.
+        If inter_point is part of all_local_coords, it is only considered one time less than it occurs.
+        inter_point is expected to be in all_local_coords at most once.
 
-        Returns: np.ndarray
-            of indices of the k nearest neighbors.
+        Parameters
+        ----------
+        all_local_coords : list
+            List of coordinates of all points.
+        inter_point : list | np.ndarray
+            Coordinates of the point for which the neighbors are to be found.
+        k : int
+            Number of neighbors to consider.
+
+        Returns
+        ------
+        neighbor_indices : np.ndarray
+            Indices of the k nearest neighbors in all local points.
         """
         assert (
-            len(all_local_coords) > k
-        ), "Desired number of neighbors must be less than the number of all available neighbors."
-        if not inter_point_is_neighbor:
-            neighbors = NearestNeighbors(n_neighbors=k + 1).fit(all_local_coords)
+            len(all_local_coords) >= k
+        ), "Desired number of neighbors must be less than or equal to the number of all available neighbors."
+        # If the number of neighbors is larger than the number of all available neighbors, increase the number of neighbors
+        # to be able to delete a neighbor if it coincides with the interpolation point.
+        if len(all_local_coords) > k:
+            k += 1
+        neighbors = NearestNeighbors(n_neighbors=k).fit(all_local_coords)
 
-            dists, neighbor_indices = neighbors.kneighbors(
-                [inter_point], return_distance=True
-            )
+        dists, neighbor_indices = neighbors.kneighbors(
+            [inter_point], return_distance=True
+        )
 
-            if np.min(dists) < 1e-10:
-                argmin = np.argmin(dists)
-                neighbor_indices = np.delete(neighbor_indices, argmin)
-            else:
-                argmax = np.argmax(dists)
-                neighbor_indices = np.delete(neighbor_indices, argmax)
+        # Check whether the inter_point is also part of the neighbor list and remove it.
+        if np.min(dists) < 1e-16:
+            argmin = np.argmin(dists)
+            neighbor_indices = np.delete(neighbor_indices, argmin)
+        # If point itself is not in neighbor list, remove neighbor with largest distance
+        # to return the desired number of neighbors
         else:
-            neighbors = NearestNeighbors(n_neighbors=k).fit(all_local_coords)
-            neighbor_indices = neighbors.kneighbors(
-                [inter_point], return_distance=False
-            )
+            argmax = np.argmax(dists)
+            neighbor_indices = np.delete(neighbor_indices, argmax)
 
         return neighbor_indices
 
     def inv_dist_weighted_interp(
-        self, neighbors: list, point, values: list
-    ) -> np.ndarray:
+        self, neighbors: np.ndarray, point: np.ndarray, values
+    ):
         """
             Interpolate a value at a point using inverse distance weighting.
+            .. math::
+                f(x) = (\sum_{i=1}^{n} \frac{f_i}{\Vert x_i - x \Vert^2}) / (\sum_{j=1}^{n} \frac{1}{\Vert x_j - x \Vert^2})
 
-        Args:
-            neighbors: list
-                Coordinates at which the values are known.
-            point:
-                Coordinates at which the value is to be interpolated.
-            values: list
-                Values at the known coordinates.
+        Parameters
+        ----------
+        neighbors : np.ndarray
+            Coordinates at which the values are known.
+        point : np.ndarray
+            Coordinates at which the value is to be interpolated.
+        values :
+            Values at the known coordinates.
 
-        Returns: nd.array
+        Returns
+        -------
+        interpol_val / summed_weights :
             Value at interpolation point.
         """
         interpol_val = 0
         summed_weights = 0
+        # iterate over all neighbors
         for inx in range(len(neighbors)):
+            # compute the squared norm of the difference between interpolation point and neighbor
             norm = np.linalg.norm(np.array(neighbors[inx]) - np.array(point)) ** 2
-            if norm < 1e-10:
+            # If interpolation point is already part of the data it is returned as the interpolation result
+            # This avoids division by zero
+            if norm < 1e-16:
                 return values[inx]
+            # update interpolation value
             interpol_val += values[inx] / norm
+            # extend normalization factor
             summed_weights += 1 / norm
 
         return interpol_val / summed_weights
--- a/micro_manager/interpolation.py
+++ b/micro_manager/interpolation.py
@@ -46,9 +46,7 @@ class Interpolation:
 
         return neighbor_indices
 
-    def inv_dist_weighted_interp(
-        self, neighbors: np.ndarray, point: np.ndarray, values
-    ):
+    def interpolate(self, neighbors: np.ndarray, point: np.ndarray, values):
         """
             Interpolate a value at a point using inverse distance weighting.
             .. math::
--- a/micro_manager/interpolation.py
+++ b/micro_manager/interpolation.py
@@ -14,7 +14,7 @@ class Interpolation:
         k: int,
     ) -> np.ndarray:
         """
-        Get the indices of the k nearest neighbors of a point in a list of coordinates.
+        Get local indices of the k nearest neighbors of a point.
 
         Parameters
         ----------
@@ -28,12 +28,12 @@ class Interpolation:
         Returns
         ------
         neighbor_indices : np.ndarray
-            Indices of the k nearest neighbors in all local points.
+            Local indices of the k nearest neighbors in all local points.
         """
         assert len(all_local_coords) > 0, "No local coordinates provided."
         if len(all_local_coords) < k:
             self._logger.info(
-                "Number of desired neighbors k = {} is larger than the number of available neighbors {}. Setting k = {}.".format(
+                "Number of desired neighbors k = {} is larger than the number of available neighbors {}. Resetting k = {}.".format(
                     k, len(all_local_coords), len(all_local_coords)
                 )
             )
@@ -48,7 +48,7 @@ class Interpolation:
 
     def interpolate(self, neighbors: np.ndarray, point: np.ndarray, values):
         """
-            Interpolate a value at a point using inverse distance weighting.
+            Interpolate a value at a point using inverse distance weighting. (https://en.wikipedia.org/wiki/Inverse_distance_weighting)
             .. math::
                 f(x) = (\sum_{i=1}^{n} \frac{f_i}{\Vert x_i - x \Vert^2}) / (\sum_{j=1}^{n} \frac{1}{\Vert x_j - x \Vert^2})
 
--- a/micro_manager/interpolation.py
+++ b/micro_manager/interpolation.py
@@ -1,0 +1,43 @@
+import numpy as np
+from scipy.interpolate import griddata as gd
+
+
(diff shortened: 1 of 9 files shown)


micro_manager/micro_manager.py
```diff
@@ -71,8 +71,6 @@ def __init__(self, config_file: str) -> None:
             self._rank,
             self._size)
 
-        micro_file_name = self._config.get_micro_file_name()
-
```
IshaanDesai, line 75: Is this related to crash handling?

micro_manager/micro_manager.py
```diff
@@ -437,13 +438,45 @@ def _solve_micro_simulations(self, micro_sims_input: list) -> list:
         micro_sims_output = [None] * self._local_number_of_sims
 
         for count, sim in enumerate(self._micro_sims):
-            start_time = time.time()
-            micro_sims_output[count] = sim.solve(micro_sims_input[count], self._dt)
-            end_time = time.time()
 
-            if self._is_micro_solve_time_required:
+            if not self._crashed_sims[count]:
+                try:
```
IshaanDesai, line 443: Please add comments on every condition or switch case that you write, so that it is easier to understand what the code is doing.

micro_manager/micro_manager.py
```diff
@@ -437,13 +438,45 @@ def _solve_micro_simulations(self, micro_sims_input: list) -> list:
...
+                    start_time = time.time()
+                    micro_sims_output[count] = sim.solve(
+                        micro_sims_input[count], self._dt)
+                    end_time = time.time()
+                except Exception as error_message:
+                    _, mesh_vertex_coords = self._participant.get_mesh_vertex_ids_and_coordinates(
+                        self._macro_mesh_name)
```
IshaanDesai, line 450: Rather than having the API call here, we can just save the coordinates and sample them as and when necessary.

micro_manager/micro_manager.py
```diff
@@ -437,13 +438,45 @@ def _solve_micro_simulations(self, micro_sims_input: list) -> list:
...
+                    self._logger.error(error_message)
+                    micro_sims_output[count] = self._old_micro_sims_output[count]
+                    self._crashed_sims[count] = True
+            else:
+                micro_sims_output[count] = self._old_micro_sims_output[count]
+            if self._is_micro_solve_time_required and not self._crashed_sims[count]:
                 micro_sims_output[count]["micro_sim_time"] = end_time - start_time
 
+            crash_ratio = np.sum(self._crashed_sims) / len(self._crashed_sims)
+            if crash_ratio > 0.2:
```
IshaanDesai, line 464: Comments would help here. In the future we could potentially make this a user configurable parameter.

micro_manager/micro_manager.py
```diff
@@ -437,13 +438,45 @@ def _solve_micro_simulations(self, micro_sims_input: list) -> list:
...
+                self._logger.info("More than 20% of the micro simulations on rank {} have crashed. "
+                                  "Exiting simulation.".format(self._rank))
```
IshaanDesai, line 466: Lets calculate the percentage number in the comment with the ratio.

IshaanDesai (changes requested):
Good first steps, but we need to do a bit more work before merging this. Please look at the comments below. A general comment is writing comments in the code such that it is readable. I will do another round of reviewing soon.

micro_manager/micro_manager.py
```diff
@@ -437,13 +438,45 @@ def _solve_micro_simulations(self, micro_sims_input: list) -> list:
...
+                sys.exit()
+
+        set_sims = np.where(micro_sims_output)
+        none_mask = np.array([item is None for item in micro_sims_output])
+        unset_sims = np.where(none_mask)[0]
+
+        for unset_sims in unset_sims:
+            self._logger.info("Micro Sim {} has previously not run. "
+                              "It will be replace with the output of the first "
+                              "micro sim that ran {}".format(unset_sims, set_sims[0][0]))
```
IshaanDesai, line 476:
```suggestion
            self._logger.info("Micro simulation {} has has crashed in the very first run attempt. "
                              "The output of the first micro sim that ran ({}) will be used as its output.".format(unset_sims, set_sims[0][0]))
```

micro_manager/micro_manager.py
tjwsch: No this is a formatting error. I will revert this and open a separate pull request with changes.

tjwsch:
The latest commit adds interpolation. Besides lacking extrapolation, scipy's griddata module also cannot interpolate on linearly dependent data, thus data on the dummy problems cannot be interpolated.  more thought must be put into the issue, to interpolate and extrapolate all cases of simulation crashes, and implementing a custom solution for interpolation is most likely required.

IshaanDesai:
Further discussions led to the idea that implementing an [inverse distance weighing](https://en.wikipedia.org/wiki/Inverse_distance_weighting) would make sense for the interpolation. This would be done with a k-nearest-neighbor method, where the user can set k.

micro_manager/interpolation.py
```diff
@@ -0,0 +1,83 @@
...
+
+class Interpolation:
+    def __init__(self, logger):
+
+        self._logger = logger
+
+    def get_nearest_neighbor_indices_local(
+        self,
+        all_local_coords,
+        inter_point,
```
IshaanDesai, line 13: Data type of input arguments are missing.

micro_manager/interpolation.py
```diff
@@ -0,0 +1,83 @@
...
+            inter_point:
+                Coordinates of the point for which the neighbors are to be found.
+            k: int
+            inter_point_is_neighbor: bool, optional
+                Decide whether the interpolation point is considered as its own neighbor.
+                Defaults to False.
+
+        Returns: np.ndarray
+            of indices of the k nearest neighbors.
+        """
```
IshaanDesai, line 32: Stick to [NumPy style docstrings](https://numpydoc.readthedocs.io/en/latest/format.html).

micro_manager/interpolation.py
```diff
@@ -0,0 +1,83 @@
...
+        assert (
+            len(all_local_coords) > k
+        ), "Number of neighbors must be less than the number of all available neighbors."
```
IshaanDesai, line 35:
```suggestion
        ), "Desired number of neighbors must be less than the number of all available neighbors."
```

IshaanDesai, line 35:
```suggestion
        ), "Desired number of neighbors must be less than the number of all available neighbors."
```

micro_manager/interpolation.py
```diff
@@ -0,0 +1,83 @@
...
+        Note: It can be chosen whether the point itself is considered as a neighbor or not.
+        Args:
+            all_local_coords: list
+                List of coordinates of all points.
+            inter_point:
+                Coordinates of the point for which the neighbors are to be found.
+            k: int
+            inter_point_is_neighbor: bool, optional
+                Decide whether the interpolation point is considered as its own neighbor.
+                Defaults to False.
```
IshaanDesai, line 28: What is the reasoning behind providing this option? Does it change the accuracy of the interpolation?

micro_manager/interpolation.py
```diff
@@ -0,0 +1,83 @@
...
+            Value at interpolation point.
+        """
+        interpol_val = 0
+        summed_weights = 0
+        for inx in range(len(neighbors)):
+            norm = np.linalg.norm(np.array(neighbors[inx]) - np.array(point)) ** 2
+            if norm < 1e-10:
+                return values[inx]
+            interpol_val += values[inx] / norm
+            summed_weights += 1 / norm
```
IshaanDesai, line 81: Add brief comments on what is done in each line, as right now it is hard to understand which operations are being done.

micro_manager/interpolation.py
```diff
@@ -0,0 +1,83 @@
...
+        Args:
+            neighbors: list
+                Coordinates at which the values are known.
+            point:
+                Coordinates at which the value is to be interpolated.
+            values: list
+                Values at the known coordinates.
+
+        Returns: nd.array
+            Value at interpolation point.
```
IshaanDesai, line 72: Does the return value have to be of the type of `nd.array`?

micro_manager/micro_manager.py
```diff
@@ -331,11 +351,11 @@ def _initialize(self) -> None:
 
         (
             self._mesh_vertex_ids,
-            mesh_vertex_coords,
+            self._mesh_vertex_coords,
```
IshaanDesai, line 354: Is this change related to this PR?

micro_manager/interpolation.py
tjwsch: If the distance is zero, inverse distance weighting will have a zero division but there is also no reason to interpolate as information at the interpolation point exists. In the case of crash interpolation, this case should never occur as there is no information and that's why interpolation is performed. The option can be removed.

micro_manager/micro_manager.py
tjwsch: Yes,  `self._mesh_vertex_coords` is used a few times in the crash handling to log where the simulation crashed and it is used for the interpolation.

micro_manager/interpolation.py
tjwsch: No, it does not

micro_manager/micro_manager.py
```diff
@@ -606,20 +702,87 @@ def _solve_micro_simulations_with_adaptivity(
...
+                        ] = self._interpolation.inv_dist_weighted_interp(
+                            coord, self._mesh_vertex_coords[unset_sim], key_values
+                        )
+                    micro_sims_output[unset_sim] = output_interpol
+                    if self._is_micro_solve_time_required:
+                        micro_sims_output[unset_sim]["micro_sim_time"] = 0
+                    micro_sims_output[unset_sim]["active_state"] = 1
+                    micro_sims_output[unset_sim][
+                        "active_steps"
+                    ] = self._micro_sims_active_steps[unset_sim]
```
IshaanDesai, line 785: We can probably move this into a function and call it once to keep the code clean. Right now we have too many if condition and loop going around, which makes it extremely hard to read code. In general up to 3 levels of loops are fine, more are a problem.

IshaanDesai (changes requested):
Good effort already :+1: my suggestions are mainly about code styling. Let us test this functionality on a real example and check if crashes are caught.

micro_manager/micro_manager.py
IshaanDesai: Okay got it :+1:

micro_manager/interpolation.py
```diff
@@ -0,0 +1,84 @@
+import numpy as np
+from sklearn.neighbors import NearestNeighbors
+
+
+class Interpolation:
+    def __init__(self, logger):
+
+        self._logger = logger
+
+    def get_nearest_neighbor_indices_local(
```
IshaanDesai, line 10:
```suggestion
    def get_nearest_neighbor_indices(
```
Unless we have a global variant of the same function, no need to specify `local` in the function name. We can just go with a generic function name and specify the local indices part in the docstring.

micro_manager/interpolation.py
```diff
@@ -0,0 +1,84 @@
...
+        self,
+        all_local_coords: np.ndarray,
+        inter_point: np.ndarray,
+        k: int,
+    ) -> np.ndarray:
+        """
+        Get the indices of the k nearest neighbors of a point in a list of coordinates.
```
IshaanDesai, line 17:
```suggestion
        Get local indices of the k nearest neighbors of a point.
```

micro_manager/interpolation.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class Interpolation:
+    def __init__(self, logger):
+
+        self._logger = logger
+
+    def get_nearest_neighbor_indices_local(
+        self,
+        all_local_coords: np.ndarray,
```
IshaanDesai, line 12:
```suggestion
        all_coords: np.ndarray,
```

micro_manager/interpolation.py
```diff
@@ -0,0 +1,84 @@
...
+        inter_point: np.ndarray,
+        k: int,
+    ) -> np.ndarray:
+        """
+        Get the indices of the k nearest neighbors of a point in a list of coordinates.
+
+        Parameters
+        ----------
+        all_local_coords : list
```
IshaanDesai, line 21:
```suggestion
        all_coords : list
```

micro_manager/interpolation.py
```diff
@@ -0,0 +1,84 @@
...
+            List of coordinates of all points.
+        inter_point : list | np.ndarray
+            Coordinates of the point for which the neighbors are to be found.
+        k : int
+            Number of neighbors to consider.
+
+        Returns
+        ------
+        neighbor_indices : np.ndarray
+            Indices of the k nearest neighbors in all local points.
```
IshaanDesai, line 31:
```suggestion
            Local indices of the k nearest neighbors in all local points.
```

micro_manager/interpolation.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        assert len(all_local_coords) > 0, "No local coordinates provided."
```
IshaanDesai, line 33: Please do a double check if this assertion makes sense here. If we never hit this case in the `micro_manager.py`, we do not need the check. Such checks typically only make sense for user-side API functions.

micro_manager/interpolation.py
```diff
@@ -0,0 +1,84 @@
...
+        if len(all_local_coords) < k:
+            self._logger.info(
+                "Number of desired neighbors k = {} is larger than the number of available neighbors {}. Setting k = {}.".format(
```
IshaanDesai, line 36:
```suggestion
                "Number of desired neighbors k = {} is larger than the number of available neighbors {}. Resetting k = {}.".format(
```

micro_manager/interpolation.py
```diff
@@ -0,0 +1,84 @@
...
+
+        neighbor_indices = neighbors.kneighbors(
+            [inter_point], return_distance=False
+        ).flatten()
+
+        return neighbor_indices
+
+    def interpolate(self, neighbors: np.ndarray, point: np.ndarray, values):
+        """
+            Interpolate a value at a point using inverse distance weighting.
```
IshaanDesai, line 51:
```suggestion
            Interpolate a value at a point using inverse distance weighting. (https://en.wikipedia.org/wiki/Inverse_distance_weighting)
```

micro_manager/micro_manager.py
```diff
@@ -551,12 +580,48 @@ def _solve_micro_simulations(self, micro_sims_input: list) -> list:
...
+                        "Micro simulation at macro coordinates {} has experienced an error. "
+                        "See next entry on this rank for error message.".format(
+                            self._mesh_vertex_coords[count]
+                        )
+                    )
+                    self._logger.error(error_message)
+                    self._has_sim_crashed[count] = True
+
+        # Interpolate result for crashed simulation
+        if self._has_sim_crashed.count(True) > 0:
```
IshaanDesai, line 610: Is this condition a must? If there are no crashed simulations, I would expect the logic in the next step to yield `unset_sims = []`.

micro_manager/micro_manager.py
```diff
@@ -654,6 +755,110 @@ def _solve_micro_simulations_with_adaptivity(
...
+            Result of the interpolation in which keys are names of data and the values are the data.
+        """
+        # Find neighbors of the crashed simulation in active and non-crashed simulations
+        # Set iteration length to only iterate over active simulations
+        if self._is_adaptivity_on:
+            iter_length = active_sim_ids
+        else:
+            iter_length = range(len(micro_sims_input))
+        micro_sims_active_input_lists = []  # DECLARATION
+        micro_sims_active_values = []  # DECLARATION
```
IshaanDesai, line 792:
```suggestion
        micro_sims_active_input_lists = []
        micro_sims_active_values = []
```
The `DECLARATION` tag is used for member variables and not local variables within a function.

tests/unit/test_micro_simulation_crash_handling.py
```diff
@@ -0,0 +1,125 @@
...
+        return {
+            "micro-vector-data": macro_data["macro-vector-data"],
+            "micro-scalar-data": macro_data["macro-scalar-data"],
+        }
+
+
+class TestSimulationCrashHandling(TestCase):
+    def test_crash_handling(self):
+        """
+        Test if the micro manager catches a simulation crash and handles it adequately.
```
IshaanDesai, line 29:
```suggestion
        Test if the Micro Manager catches a simulation crash and handles it adequately.
```

IshaanDesai (commented):
We are getting closer to merging this :grin: some more comments on my side. The logic and overall code looks good now.

micro_manager/interpolation.py
tjwsch: Yes, this is redundant as this is ensured within the micro-manager.

micro_manager/micro_manager.py
tjwsch: No, it is redundant.

IshaanDesai (approved):
Good to go :+1: please remember to **squash and merge**.

PR #93, opened 2024-04-09
Add switch to turn adaptivity on and off in configuration

IshaanDesai:
A boolean switch is added to turn adaptivity on and off in the configuration, because JSON does not allow comments. A user who has defined adaptivity settings is forced to manually delete the lines if they want to run a simulation without adaptivity.


micro_manager/config.py
```diff
@@ -64,122 +64,166 @@ def read_json(self, config_filename):
...
-                "Domain decomposition is not specified, so the Micro Manager will expect to be run in serial.")
+                "Domain decomposition is not specified, so the Micro Manager will expect to be run in serial."
+            )
 
         try:
-            if data["simulation_params"]["adaptivity"]:
+            if data["simulation_params"]["adaptivity"] == "True":
                 self._adaptivity = True
             else:
                 self._adaptivity = False
```
Fujikawas, line 130: Could also warn the case when the adaptivity parameters are given but adaptivity is set to false.

micro_manager/config.py
```diff
@@ -64,122 +64,166 @@ def read_json(self, config_filename):
...
         except BaseException:
             self._logger.info(
-                "Micro Manager will not adaptively run micro simulations, but instead will run all micro simulations in all time steps.")
+                "Micro Manager will not adaptively run micro simulations, but instead will run all micro simulations in all time steps."
```
Fujikawas, line 133: "No adaptivity settings are given, Micro Manager will run all micro simulations in all time steps instead of adaptively run them."

Fujikawas (approved):

PR #94, opened 2024-04-09
Fix doc image prefix

fsimonis:
This PR fixes the doc image prefix and adds a pre-commit that ensures correct prefixes for future images.


IshaanDesai (approved):
Thanks!

PR #96, opened 2024-04-22
TBD: Follow time stepping proposed in documentation

BenjaminRodenberg:
In the step-by-step guide on precice.org it is suggested to get the maximum allowed time step size from preCICE before calling advance.

To me it looks dangerous how things are done at the moment, because actually the following is happening:

```
self._participant.advance(self._dt)
self._dt = self._participant.get_max_time_step_size()

t += self._dt
n += 1
```
We tell preCICE "I just finished window `n` with time step size `self._dt`, but then update the internal time `t` by the step size of the **upcoming** window. I always prefer to first do all the updates and then call advance:

```
self._dt = self._participant.get_max_time_step_size()    # assuming this has been called at the beginning of the current time step
...
# time step is finished, update internal counters
t += self._dt 
n += 1

self._participant.advance(self._dt)  # tell preCICE: I'm done and ready to go to the next time step / iteration
```
This should generally also work with subcycling and/or the participant-first method (no guarantee, because this needs testing and is always easy to break, if hard-to-spot errors are in the code)

### DIFF
--- a/micro_manager/micro_manager.py
+++ b/micro_manager/micro_manager.py
@@ -173,6 +173,9 @@ class MicroManager:
                 )
 
         while self._participant.is_coupling_ongoing():
+
+            self._dt = self._participant.get_max_time_step_size()  # ask preCICE at beginning of time step for allowed time step size
+
             # Write a checkpoint
             if self._participant.requires_writing_checkpoint():
                 for i in range(self._local_number_of_sims):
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
The comments I put into the code are a bit excessive. Feel free to remove them. They are mainly intended for simplifying the review process.

IshaanDesai (approved):
Seems logical if one goes by the latest documentation on the preCICE website.

Issue #89, opened 2024-03-28
Add functionality for spanshot computation from a full-order model


IshaanDesai:
In the framework of the [SimTech project PN5-9(II)](https://www.simtech.uni-stuttgart.de/exc/research/pn/pn5/pn5-9ii/) the Micro Manager needs to be extended to be able to do an offline snapshot computation step for a given full-order model. The input parameter space for creating the snapshots will be defined by the user through a configuration file. The existing JSON file could be potentially used for this, or a new configuration file format can be explored. The data required from the simulations (output) is also defined by the user. The database developed from this computation should have an interface between its output and the reduced-order/surrogate models, presumably using [hdf5](https://github.com/HDFGroup/hdf5). Rough tasks to add this functionality are:

- [x] Design configuration options for parameter space and data output options to run a full-order model
- [x] Finalize a file format for configuring the parameters of the snapshot computation
- [x] Implement a new class in the Micro Manager to run the offline computation
- [x] Find the most suitable file format to store the snapshots and create an interface to surrogate/reduced-order models
- [x] For a simple dummy simulation, try to output the snapshots in the file format finalized in the previous step

uekerman:
And (advanced)
- [ ] Provide a callback interface to call any user-provided (Python?) post-processing script

IshaanDesai:
Closed via https://github.com/precice/micro-manager/pull/101

PR #101, opened 2024-05-13
Snapshot computation

tjwsch:
Partly implementation of snapshot features discussed in #89. This does not yet include the configuration of the parameter space at runtime and a callback interface to invoke postprocessing as discussed in [SimTech project PN5-9(II)](https://www.simtech.uni-stuttgart.de/exc/research/pn/pn5/pn5-9ii/). 
Reading parameter information from an hdf5 file and writing snapshot outputs to hdf5 files and the project structure might need more work.


examples/snapshot_example/snapshot-config.json
```diff
@@ -0,0 +1,13 @@
+{
+    "micro_file_name": "micro_dummy",
+    "snapshot_params": {
+        "parameter_file_name": "parameter.hdf5",
+        "read_data_names": {"macro-scalar-data": "scalar", "macro-vector-data": "vector"},
+        "write_data_names": {"micro-scalar-data": "scalar", "micro-vector-data": "vector"},
+        "postprocessing": "True",
```
tjwsch, line 7: Instead of having a boolean to activate post-processing and post-processing as part of the micro-simulation class, we want to provide the name of a separate post-processing script. This allows removing the snapshot example and reusing the Micro Manager example micro simulation classes by adding the files necessary to run snapshot computation.

snapshot/read_write_hdf.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+
+        parameter_file = h5py.File(file_path, "r")
+        parameter_data = dict()
+        output = []
+        # Read data by iterating over the relevant datasets
+        for key in data_names.keys():
+            parameter_data[key] = np.asarray(parameter_file[key][:])
+            my_key = key
+        # 3. iterate over len of data. In each iteration write data from all macro data sets (using the read_data_names as keys) to a dictionary with the key being the name of the data set
```
tjwsch, line 122: Review comments.

snapshot/read_write_hdf.py
```diff
@@ -0,0 +1,130 @@
...
+        # and the value being the current value.
+        for i in range(len(parameter_data[my_key])):
+            current_data = dict()
+            for key in data_names.keys():
+                current_data[key] = parameter_data[key][i]
+            output.append(current_data)
+        # 4. return the dictionary list
```
tjwsch, line 129: Remove comment

snapshot/snapshot_computation.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        Solve the problem by iterating over a set macro parameters.
+        - Read macro parameters from the config file, solve micro simulation one by one,
+          and write outputs to a file.
+        """
+
+        # Loop over all macro parameters
+        for elems in range(self._local_number_of_sims):
+            # Create micro simulation object
+            self._micro_sims = create_simulation_class(self._micro_problem)(
```
tjwsch, line 100: In contrast to the Micro Manager, micro simulations are newly initialized in the `solve()` method as each microsimulation is only required for one loop of the `solve()` method. Is it a good solution to initalize microsimulations within the solve call as this is technically part of the `_initialize()`-method's responsibility?

examples/snapshot_example/.gitignore
tjwsch, line 1: Remove as a result of the re-structuring of the post-processing. Move `output/` to example .gitignore.

examples/snapshot_example/README.md
tjwsch, line 1: Remove as a result of the re-structuring of the post-processing. Add the content to README.md of examples and distinguish between running coupling and snapshot computation

examples/snapshot_example/cpp-dummy/.gitignore
tjwsch, line 1: Remove as a result of the re-structuring of the post-processing.

examples/snapshot_example/cpp-dummy/micro_cpp_dummy.cpp
tjwsch, line 1: Remove as a result of the re-structuring of the post-processing.

examples/snapshot_example/cpp-dummy/micro_cpp_dummy.hpp
tjwsch, line 1: Remove as a result of the re-structuring of the post-processing.

examples/snapshot_example/cpp-dummy/run_snapshot.py
tjwsch, line 1: Move to examples as a result of the restructuring of the post-processing.

examples/snapshot_example/python_dummy/.gitignore
tjwsch, line 1: Remove as a result of the restructuring of the post-processing.

examples/snapshot_example/python_dummy/micro_dummy.py
tjwsch, line 1: Remove as a result of the restructuring of the post-processing.

examples/snapshot_example/python_dummy/run_snapshot.py
tjwsch, line 1: Move to examples as a result of the restructuring of the post-processing.

examples/snapshot_example/snapshot-config.json
tjwsch, line 1: Move to examples as a result of the restructuring of the post-processing.

snapshot/snapshot_computation.py
```diff
@@ -0,0 +1,305 @@
...
+                        getattr(self._micro_problem, "postprocessing")
+                    ):
+                        micro_sims_output = self._micro_sims.postprocessing(
+                            micro_sims_output
+                        )
+                    else:
+                        self._logger.info(
+                            "Postprocessing is activated in config file but not available. Skipping postprocessing."
+                        )
+                        self._is_postprocessing_required = False
```
tjwsch, line 122: Restructure to execute an additional post-processing script.

snapshot/snapshot_computation.py
```diff
@@ -0,0 +1,305 @@
...
+            Dict in which keys are names of data and the values are the data which are required inputs to
+            solve a micro simulation.
+
+        Returns
+        -------
+        micro_sims_output : dict | None
+            Dicts in which keys are names of data and the values are the data of the output of the micro
+            simulations. The return type is None if the simulation has crashed.
+        """
+        # If time is a parameter, split input as time is a separate argument
```
tjwsch, line 260: What is the best way to handle `dt`? If micro simulations require `dt` should it be a macro parameter, to allow it to change over different microsimulations or a fixed value specified in the config file?  What should `dt` be set to, if a microsimulation is independent of `dt`, e.g. the micro dummies?

tjwsch (commented):
There are some open questions and restructuring of postprocessing is required.

examples/snapshot_example/snapshot-config.json
IshaanDesai: Good idea! Let us try this out after the feature itself is close to being merged.

snapshot/config_snapshot.py
```diff
@@ -0,0 +1,228 @@
...
+
+        self._merge_output = False
+
+        self._output_micro_sim_time = False
+
+        self.read_json(config_filename)
+
+    def read_json(self, config_filename):
+        """
+        Reads JSON adapter configuration file and saves the data to the respective instance attributes.
```
IshaanDesai, line 44:
```suggestion
        Reads JSON snapshot configuration file and saves the data to the respective instance attributes.
```

snapshot/config_snapshot.py
```diff
@@ -0,0 +1,228 @@
...
+        Depending on user input, Snapshot computation will calculate execution time of solve() step of every micro simulation.
+
+        Returns
+        -------
+        output_micro_sim_time : bool
+            True if micro simulation solve time is required.
+        """
+        return self._output_micro_sim_time
+
+    def get_postprocessing(self):
```
IshaanDesai, line 208: This is essentially a boolean, right? We should have a more intuitive name of the function.

snapshot/read_write_hdf.py
```diff
@@ -0,0 +1,130 @@
...
+import os
+
+
+class ReadWriteHDF:
+    def __init__(self, logger) -> None:
+        self._logger = logger
+
+    def create_file(self, file_path: str) -> None:
+        """
+        Create a file with a given name in a given directory overriding a file if it exists.
```
IshaanDesai, line 13:
```suggestion
        Create a file with a given name in a given directory.
        Overwriting a file if it exists.
```

snapshot/read_write_hdf.py
```diff
@@ -0,0 +1,130 @@
...
+                    # If data set does not exist create and add current data
+                    if key not in main_file.keys():
+                        main_file.create_dataset(
+                            key,
+                            data=current_data[i],
+                            shape=(1, *current_data[i].shape),
+                            maxshape=(None, *current_data[i].shape),
+                        )
+                    # If dataset exists resize and add current data
+                    else:
```
IshaanDesai, line 55:
```suggestion
                    else: # If dataset exists, resize and add current data
```

snapshot/read_write_hdf.py
```diff
@@ -0,0 +1,130 @@
...
+        ----------
+        file_path : str
+            Path to file in which the data should be written.
+        macro_data : dict
+            Dict of macro simulation input.
+        micro_data : dict
+            Dict of micro simulation output.
+        """
+        parameter_file = h5py.File(file_path, "a")
+        input_data = [macro_data, micro_data]
```
IshaanDesai, line 77: Is this a way to merge two dicts into a new dict? Very nice!

snapshot/read_write_hdf.py
```diff
@@ -0,0 +1,130 @@
...
+                if key not in parameter_file.keys():
+                    parameter_file.create_dataset(
+                        key,
+                        data=current_data,
+                        shape=(1, *current_data.shape),
+                        maxshape=(None, *current_data.shape),
+                    )
+                    parameter_file[key]
+                # If dataset exists resize and add current data
+                else:
```
IshaanDesai, line 93:
```suggestion
                else: # If dataset exists, resize and add current data
```

snapshot/read_write_hdf.py
IshaanDesai: Any input required here? Shortening the comment would be good.

snapshot/snapshot_computation.py
```diff
@@ -0,0 +1,305 @@
...
+import time
+
+from .config_snapshot import SnapshotConfig
+from .read_write_hdf import ReadWriteHDF as hp
+from micro_manager.micro_simulation import create_simulation_class
+
+sys.path.append(os.getcwd())
+
+
+class SnapshotComputation:
```
IshaanDesai, line 31: Can we make this a child class of the MicroManager class? We redefine many member variables anyway. Perhaps this would make the overall implementation of the snapshot computation cleaner. The justification for doing this is that the snapshot computation is in a way a use case of the Micro Manager, but for a different purpose. So we are in fact running the Micro Manager, but just in a different way.

snapshot/snapshot_computation.py
IshaanDesai: We need to be careful with the terms here. The step here is the **creation** of the micro simulations, and not their initialization. Initialization is essentially optional, and only happens if the micro simulation library has a function called `initialize()`. Initialization of the micro simulations is handled in the `initialize()` function of the MicroManager class. Inheriting from this class and calling this function would make sense.

snapshot/snapshot_computation.py
```diff
@@ -0,0 +1,305 @@
...
+
+        # Loop over all macro parameters
+        for elems in range(self._local_number_of_sims):
+            # Create micro simulation object
+            self._micro_sims = create_simulation_class(self._micro_problem)(
+                self._global_ids_of_local_sims[elems]
+            )
+
+            micro_sims_input = self._macro_parameters[elems]
+            # Solve micro simulation
```
IshaanDesai, line 105:
```suggestion
            # Solve micro simulations
```

snapshot/snapshot_computation.py
```diff
@@ -0,0 +1,305 @@
...
+                for key, value in micro_sims_input.items():
+                    parameter += "{} = {}, ".format(key, value)
+                parameter = parameter[:-2]
+                self._logger.info(
+                    "Skipping snapshot storage for crashed simulation with parameter {}.".format(
+                        parameter
+                    )
+                )
+
+        # If activated, merge output files
```
IshaanDesai, line 138: If what is activated?

snapshot/snapshot_computation.py
```diff
@@ -0,0 +1,305 @@
...
+    # Private methods
+    # ***************
+
+    def _initialize(self) -> None:
+        """
+        Initialize the Snapshot Computation by performing the following tasks:
+        - Distribute the parameter data equally if the snapshot creation is executed in parallel.
+        - Read macro parameter from parameter file.
+        - Create output subdirectory and file paths to store output.
+        - Import micro simulation class.
```
IshaanDesai, line 161: Please follow the docstring formatting.

snapshot/snapshot_computation.py
```diff
@@ -0,0 +1,305 @@
...
+import importlib
+import os
+import sys
+from mpi4py import MPI
+import numpy as np
+import logging
+import time
+
+from .config_snapshot import SnapshotConfig
+from .read_write_hdf import ReadWriteHDF as hp
```
IshaanDesai, line 25: Lets have a more elaborate abbreviation than just `hp`.

IshaanDesai (changes requested):
Good work already :clap: My suggestions are fundamental and hence require some discussion before large implementation changes. Let us discuss here in the comments.

tests/unit/test_snapshot_computation.py
```diff
@@ -0,0 +1,154 @@
+import numpy as np
+import os
+from unittest import TestCase
+from unittest.mock import MagicMock, patch
+import snapshot
```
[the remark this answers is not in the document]
tjwsch: I agree. We should make the snapshot computation part of the Micro Manager package. When I moved the snapshot computation to a new directory, I wasn't aware of that effect on the packaging.

snapshot/read_write_hdf.py
tjwsch: No, it creates a list of two dicts. But there are simple ways to merge two dicts into one and it makes sense to do that here as it would allow us to remove one of the for-loops and make the code more readable.

snapshot/snapshot_computation.py
tjwsch: I thought about inheritance but to me, there is too much difference in all the class methods to make it seem sensible. I believe it's possible to inherit from the Micro Manager but the Micro Manager also sets many member variables that are not required or available for the snapshot computation, e.g. everything related to preCICE, so I'm not sure how much of the Micro Manager could be inherited. Not using preCICE in the snapshot computation makes it difficult to reuse parts of the Micro Manager as much of the managing process depends on information from or consists of information written to preCICE. What could make it more natural to integrate the Micro Manager into the snapshot computation is to have a class that can imitate preCICE interface by performing the necessary operations for the snapshot computation and handling the snapshot database writing. This class could be used for the snapshot computation and communicate macro parameters to the snapshot computation. This approach could allow us to reuse much of the Micro Manager for the snapshot computation.

snapshot/config_snapshot.py
```diff
@@ -0,0 +1,228 @@
+"""
+Class Config provides functionality to read a JSON file and pass the values to the Snapshot computation.
+"""
+
+import json
+import os
+from warnings import warn
+
+
+class SnapshotConfig:
```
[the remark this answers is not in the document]
tjwsch: Yes, we can. There were two reasons I did it this way: 

1. making the config classes more easily readable
2. ensuring there aren't so many unset (i.e. None valued) getter-methods in the config file. 

But a few more getter methods won't make the code much harder to read (to 1.) and I realize this is already done in the Micro Manager if adaptivity is deactivated or in the sequential case (to 2.).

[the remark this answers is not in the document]
IshaanDesai: > I realize this is already done in the Micro Manager if adaptivity is deactivated or in the sequential case (to 2.).

Exactly, so lets just reuse the `Config` class.

snapshot/read_write_hdf.py
IshaanDesai: Okay. Yes, lets go for merging the two dicts into a new dict and then using that; it will simplify the logic quite a bit.

snapshot/snapshot_computation.py
IshaanDesai: > there is too much difference in all the class methods to make it seem sensible

If a function is redefined in the child class, the original function in the parent class will be ignored. So we will be replacing the functionality in a way.

Are you suggesting we mock the preCICE interface and have a dummy Micro Manager class? How would the original `MicroManager` class then function? Does it also inherit from this dummy class?

tjwsch: I thought that maybe it is possible to set `self._participant` in the snapshot child of the Micro Manager as a class within the Micro Manager package that replaces preCICE in the snapshot computation by having the necessary methods of preCICE and e.g. `read_data`  could read parameter data from the parameter file and `write_data` could write output to a file. Then it could be possible to use the Micro Manager class for most of the snapshot computation. But crash handling and domain decomposition must be handled differently and would still be complicated.

If that is not possible, I think it is possible to use the `__init__` from the Micro Manager but that `initialize()` and `solve()` are too different to use the corresponding methods from the Micro Manager.

IshaanDesai: This sounds rather complicated and over-engineered. I would suggest inheriting from the Micro Manager class, using the member variables which are relevant to the snapshot computation, and redefining the necessary functions once more. When you redefine `initialize()` and `solve()` in the `SnapshotComputation` class, the original functions will be ignored.

tjwsch: Okay, that sounds like a good compromise to include inheritance to me! I know they will be ignored. I thought it might be possible to use much of the entire Micro Manager for the snapshot computation, but I agree it would be over-engineered that way and I am not convinced it would be possible in a clean way.

IshaanDesai: Sounds good :+1: we will at least shorten the constructor and such. After we have merged the feature, we can think again about redesigning the functions in `SnapshotComputation` to somehow use functions from `MicroManager`.

tjwsch: After looking further into this, inheriting from `MicroManager` looks complicated (I probably saw these issues the first time I thought about inheritance). Member variables set in the constructor are only available if the constructor is called, but calling the `MicroManager` constructor will lead to errors (and warnings) as the `MicroManager` constructor tries to read a Micro Manager-type config file that does not coincide with the snapshot computation file and it tries to create a preCICE participant, which cannot work. I cannot find any clean way around this. So the only option(s) for inheritance I see is to let both the Micro Manager and the Snapshot Computation inherit from a common base class that handles common member variables (or re-designing the Micro Manager constructor).

snapshot/snapshot_computation.py
uekerman: @IshaanDesai Shouldn't this also be a user input for "normal" coupled simulations?  To allow subcycling of the micro sims?

tjwsch: In addition, I think not all micro simulations require a time step, i.e., have an explicit time dependence, especially mechanical problems (used for the snapshot computation). However, the Micro Manager currently requires a time step to be set, which, if we have a micro simulation in Python, we could just set to `None`  but a C++ implementation requires a float, which I don't think is optimal. The structure of the micro simulation template makes it necessary to set a time step, even if the micro simulation does not depend on it.

IshaanDesai: Currently we do not allow the micro simulations to have their own time step, but force the macro time step on the micro simulations. Subcycling makes sense, and the time step needs to be an input parameter in the coupled simulations already.

IshaanDesai: The micro time window size is being added as a parameter to the `coupling_params` group in https://github.com/precice/micro-manager/pull/112

tjwsch:
For large parameter spaces, we need to think about chunking. Everything I have found suggests that, within a dataset, all chunks must have the same size, which complicates matters. A poor choice of chunk size can lead, besides other effects, to files containing empty allocated space. Not knowing the dimensions of a snapshot also complicates finding an optimal chunk size. As subfiling is currently not supported in `h5py`, writing to one file from multiple processes is not the best idea, and activating MPI support in `h5py` requires additional steps, writing to a separate file from every process and merging the output files, seems to be the best option available with h5py at this time.
The first option would be to make the user provide the parameter space file in chunks This would allow processes to read only the chunks they require, but slightly complicate the parameter file creation process for the user, and if each file only reads whole chunks leads, to some load imbalances. If chunks are split up between processes this complicates the the decomposition. This would also pre-set a chunk size, which could have advantages and disadvantages.
The second option would be to read the parameter file as a whole and only have chunked output files. Again, the questions of optimal chunk sizes and how to minimize the empty file space are relevant. I think cases exist where each process would have empty allocated space, which would accumulate in the merged file unless the merging does not take place chunk-wise but snapshot-wise, and a new chunk size for the merged file is determined.

uekerman:
> make the user provide the parameter space file in chunks

This sounds like sth we should better hide from the user.

examples/.gitignore
```diff
@@ -6,3 +6,6 @@ precice-events/
 # PyCharm related files
 .idea
 __pycache__
+
+# Snapshot results
+output/
```
IshaanDesai, line 11: If this is folder is created due to a custom test or so, lets not put it in the gitignore. gitignore should only be for ignoring really unavoidable things.

examples/snapshot-config.json
```diff
@@ -0,0 +1,17 @@
...
+    },
+    "simulation_params": {
+        "micro_dt": 1.0
+    },
+    "snapshot_params": {
+        "post_processing_file_name": "snapshot_postprocessing"
+    },
+    "diagnostics": {
+        "output_micro_sim_solve_time": "True"
+      }
```
IshaanDesai, line 16: Are diagnostics necessary for an example to show the snapshot computation?

examples/snapshot_postprocessing.py
IshaanDesai, line 1: Please add a description on what this file is for.

micro_manager/config.py
```diff
@@ -452,6 +511,29 @@ def get_micro_dt(self):
...
+        -------
+        parameter_file_name : string
+            Name of the hdf5 file containing the macro parameters.
+        """
+
+        return self._parameter_file_name
+
+    def get_postprocessing_file_name(self):
+        """
+        Depending on user input, Snapshot computation will perform postprocessing for every micro simulation before writing output to a file.
```
IshaanDesai, line 528:
```suggestion
        Depending on user input, snapshot computation will perform postprocessing for every micro simulation before writing output to a file.
```

micro_manager/snapshot/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+    )
+
+import numpy as np
+
+try:
+    import h5py
+except ImportError:
+    raise ImportError(
+        "The Micro Manager snapshot computation requires the h5py package."
+    )
```
IshaanDesai, line 21: Why does this piece of code occur twice?

micro_manager/snapshot/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        f.attrs["status"] = "writing"
+        f.attrs["MicroManager_version"] = str(metadata.version("micro-manager-precice"))
+        f.attrs["date"] = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        f.close()
+
+    def collect_output_files(
+        self, dir_name: str, file_list: list, database_length: int
+    ) -> None:
+        """
+        Iterate over given HDF5 files in a given directory and copy the content into a single file.
```
IshaanDesai, line 49:
```suggestion
        Iterate over a list of HDF5 files in a given directory and copy the content into a single file.
```

micro_manager/snapshot/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+            List of files to be combined.
+        dataset_length : int
+            Global number of snapshots.
+        """
+        # Create a output file
+        main_file = h5py.File(os.path.join(dir_name, "snapshot_data.hdf5"), "w")
+        main_file.attrs["status"] = "writing"
+        main_file.attrs["MicroManager_version"] = str(
+            metadata.version("micro-manager-precice")
+        )
```
IshaanDesai, line 66: Does this pick the latest version of the package?

micro_manager/snapshot/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+                "crashed_snapshots", data=crashed_snapshots, dtype=int
+            )
+        main_file.attrs["status"] = "finished"
+        main_file.close()
+
+    def write_output_to_hdf(
+        self,
+        file_path: str,
+        macro_data: dict,
+        micro_data: dict | None,
```
IshaanDesai, line 114: Is this official syntax? If yes, very cool :smile:

micro_manager/snapshot/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        idx: int,
+        length: int,
+    ) -> None:
+        """
+        Write the output of a micro simulation to an HDF5 file.
```
IshaanDesai, line 119:
```suggestion
        Write the output of a micro simulation to a HDF5 file.
```

micro_manager/snapshot/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+                    chunks=(1, *current_data.shape),
+                    fillvalue=np.nan,
+                )
+            self._has_datasets = True
+
+        # Iterate over macro and micro data sets and write current simulation data to the file
+        for key in input_data.keys():
+            current_data = np.asarray(input_data[key])
+            parameter_file[key][idx] = current_data
+        parameter_file.close()
```
IshaanDesai, line 156:
```suggestion

        parameter_file.close()
```

micro_manager/snapshot/dataset.py
```diff
@@ -0,0 +1,238 @@
...
+
+        Returns
+        -------
+        int
+            Size of Parameter Space
+        """
+        with h5py.File(file_path, "r") as file:
+            return file[list(file.keys())[0]].len()
+
+    def write_crashed_snapshots_macro(self, file_path: str, crashed_input: list):
```
IshaanDesai, line 214: Is `_macro` in the name necessary here? What does it signify?

micro_manager/snapshot/snapshot.py
```diff
@@ -0,0 +1,281 @@
+#!/usr/bin/env python3
+"""
+Snapshot Computation is a tool withing the Micro Manager to initialize micro simulations and create a snapshot database by simulating micro simulations related to a set of prescribed macro parameters.
```
IshaanDesai, line 3:
```suggestion
Snapshot Computation is a tool within the Micro Manager to initialize micro simulations and create a snapshot database of theirs outputs by running them with a set of prescribed input parameters.
```

micro_manager/snapshot/snapshot.py
```diff
@@ -0,0 +1,281 @@
...
+                        )
+                        self._post_processing_file_name = None
+                self._data_storage.write_output_to_hdf(
+                    self._output_file_path,
+                    micro_sims_input,
+                    micro_sims_output,
+                    elems,
+                    len(self._macro_parameters),
+                )
+            # Log error and wrote macro data to database if simulation has crashed
```
IshaanDesai, line 105:
```suggestion
            # Log error and write macro data to database if simulation has crashed
```

micro_manager/snapshot/snapshot.py
```diff
@@ -0,0 +1,281 @@
...
+            else:
+                parameter = ""
+                for key, value in micro_sims_input.items():
+                    parameter += "{} = {}, ".format(key, value)
+                parameter = parameter[:-2]
```
IshaanDesai, line 110: What and why is this?

tests/unit/snapshot-config.json
```diff
@@ -0,0 +1,17 @@
...
+    },
+    "simulation_params": {
+        "micro_dt": 1.0
+    },
+    "snapshot_params": {
+        "post_processing_file_name": "snapshot_post_processing"
+    },
+    "diagnostics": {
+        "output_micro_sim_solve_time": "True"
+    }
```
IshaanDesai, line 16: Is this diagnostics output necessary for a test of the snapshot functionality?

IshaanDesai (commented):
Looks good :+1: this is close to merging now. Lets resolve some last comments. Please also remember that documentation for snapshot computation needs to be added. As discussed earlier, we can do this in a separate pull request.

micro_manager/snapshot/dataset.py
tjwsch: It picks the currently installed package version. If changes are made locally it will show that.

examples/snapshot-config.json
tjwsch: No, they are not necessary.

micro_manager/snapshot/dataset.py
tjwsch: That must be some rebasing error :grimacing:

micro_manager/snapshot/snapshot.py
tjwsch: This logs the input parameters for which a crash occurred providing this information to the user immediately. I guess it can be removed as it is also stored elsewhere.

tests/unit/snapshot-config.json
tjwsch: No

examples/.gitignore
tjwsch: The results of a snapshot computation are stored in the output folder. I don't want to upload the hdf5 files related to a specific simulation, which could happen accidentally. The output folder is created automatically. An alternative would be to ignore only the files within the folder.

micro_manager/snapshot/dataset.py
IshaanDesai: Okay good :+1:

IshaanDesai (approved):

PR #104, opened 2024-05-14
Optionally use initial macro data to initialize micro simulations

IshaanDesai:
This PR adds functionality to use initial data written by the macro side for initializing the micro simulations. This is done only when the `initialize()` function in the micro simulation library requires the macro data as input. The requirements of data are checked rigorously to catch possible mismatches.

### DIFF
--- a/micro_manager/micro_manager.py
+++ b/micro_manager/micro_manager.py
@@ -467,6 +467,7 @@ class MicroManager:
         else:
             is_initial_data_available = True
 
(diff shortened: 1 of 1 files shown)


IshaanDesai:
Tests for this functionality are purposely not added here, because it was concluded that the embedded `initialize()` call in the MicroManager class is problematic and needs to be removed. When the call is made explicit, testing will be simplified, and the tests for this functionality will be directly added in the new test framework.

mathiskelm (approved):
Looks good, thanks.

PR #112, opened 2024-06-10
Set initial time window size of the micro simulation in the configuration

IshaanDesai:
This PR implements changes which allow for the user to set the initial time window size (dt) of the micro simulation during configuration. The time window size is then using in the coupling, which allows for subcycling. 

**Note**: The micro simulation cannot modify the time window size during the coupling. A followup feature can change the micro simulation library API in a way that the `solve(...)` function call can return the desired time step value of the micro simulation.

Checklist:

### DIFF
--- a/micro_manager/micro_manager.py
+++ b/micro_manager/micro_manager.py
@@ -426,7 +426,7 @@ class MicroManagerCoupling(MicroManager):
         self._micro_sims_init = False  # DECLARATION
 
         # Read initial data from preCICE, if it is available
-        initial_data = self._read_data_from_precice(dt=self._micro_dt)
+        initial_data = self._read_data_from_precice(dt=0)
 
         if not initial_data:
             is_initial_data_available = False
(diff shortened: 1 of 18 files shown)


BenjaminRodenberg:
@IshaanDesai Anything specific where you would need feedback? It's a bit hard for me to give a general review here because I'm not that much involved in this project.

micro_manager/micro_manager.py
```diff
@@ -427,7 +426,7 @@ def initialize(self) -> None:
         self._micro_sims_init = False  # DECLARATION
 
         # Read initial data from preCICE, if it is available
-        initial_data = self._read_data_from_precice()
+        initial_data = self._read_data_from_precice(dt=self._micro_dt)
```
IshaanDesai, line 429: @BenjaminRodenberg what is the correct `dt` value to use when reading initial data from preCICE?

BenjaminRodenberg: `dt=0`, if you want to read the data at the beginning of the window.

PR #117, opened 2024-07-26
Add option for lazy creation of micro simulations

mathiskelm:
Micro simulations that are never solved don't require the microsim object to be created in the first place. If the creation/initialization is costly, avoiding it can save significant time/memory and allow simulation on smaller systems/of larger problems. In this PR the creation of micro simulations is delayed to activation.

TODO:
- [ ] Avoid code duplication for creation of micro sims,
- [ ] Improve compatibility with other Micro-Manager capabilities such as initialization of micro-sims.
- [ ] Improve error handling of incompatible settings (without initial macro data the simulation will start with only 1 micro sim)

Checklist:


micro_manager/config.py
```diff
@@ -517,6 +527,29 @@ def is_adaptivity_required_in_every_implicit_iteration(self):
...
+        """
+        Check if CPU time of the adaptivity computation needs to be output.
+
+        Returns
+        -------
+        adaptivity_cpu_time : bool
+            True if CPU time of the adaptivity computation needs to be output, False otherwise.
+        """
+        return self._adaptivity_output_cpu_time
+
```
mathiskelm, line 552:
```suggestion
```
mistake in resolving merge

micro_manager/micro_manager.py
```diff
@@ -426,6 +461,10 @@ def initialize(self) -> None:
         if hasattr(micro_problem, "initialize") and callable(
             getattr(micro_problem, "initialize")
         ):
+            if self._lazy_init:
+                raise Exception(
+                    "Adaptivity can't use data returned by initialize function of micro sims when using lazy initialization."
+                )
```
mathiskelm, line 467: An exception might be too strong based on whether the micro simulation defines the initialize function? Maybe one can not call it and log a warning? Then I should probably call initialize when creating the micro simulation, passing the current macro data? Not so clear.

micro_manager/config.py
IshaanDesai: This seems to have come from the dynamic load balancing branch, where it has now been removed.

micro_manager/micro_manager.py
IshaanDesai: It could happen that some internal definitions are done via the `initialize` function, so throwing an exception here is indeed a bit too strict. But on the other hand, if lazy initialization is done, your question of which macro data to be passed is an important one. For now we can pass the most recent macro data, but this could have implications which I cannot fully foresee right now.

IshaanDesai (approved):
Thanks for this contribution! It is already almost ready to merge. I have only minor comments/suggestions.

micro_manager/micro_manager.py
mathiskelm: Softened to a warning but currently `initialize` is not called when using lazy initialization. It should be called on lazy creation of micro-simulation objects. If it requires macro data, the most recent data should be fine. The micro simulations still obtains their state from the micro manager based on the previously associated micro simulation.

IshaanDesai: I will open an issue about this.

PR #118, opened 2024-07-26
Explicitly check if the time window has completed

IshaanDesai:
Currently, the Micro Manager understands that a time window has converged when the API call `requires_reading_checkpoint()` is false. However, `requires_reading_checkpoint()` may return false when the time window has not converged. This could happen, for example, when the Micro Manager is subcycling. To avoid falsely indicating convergence in the Micro Manager, an explicit check using the API command `is_time_window_complete()` is necessary.

Checklist:

### DIFF
--- a/micro_manager/micro_manager.py
+++ b/micro_manager/micro_manager.py
@@ -284,7 +284,9 @@ class MicroManagerCoupling(MicroManager):
                         is_sim_active = np.copy(is_sim_active_cp)
                         sim_is_associated_to = np.copy(sim_is_associated_to_cp)
 
-            else:  # Time window has converged, now micro output can be generated
+            if (
+                self._participant.is_time_window_complete()
+            ):  # Time window has converged, now micro output can be generated
                 self._logger.info(
                     "Micro simulations {} - {} have converged at t = {}".format(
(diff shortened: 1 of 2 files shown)


micro_manager/micro_manager.py
```diff
@@ -284,7 +284,9 @@ def solve(self) -> None:
                         is_sim_active = np.copy(is_sim_active_cp)
                         sim_is_associated_to = np.copy(sim_is_associated_to_cp)
 
-            else:  # Time window has converged, now micro output can be generated
+            if (
+                self._participant.is_time_window_complete()
+            ):  # Time window has converged, now micro output can be generated
```
BenjaminRodenberg, line 289:
```suggestion
            # Time window has converged, now micro output can be generated
            if self._participant.is_time_window_complete():
```
Formatting looks a bit odd.

BenjaminRodenberg (approved):
Looks ok. Before there was clearly something wrong if you are using subcycling.

micro_manager/micro_manager.py
IshaanDesai: I leave the formatting to [black](https://github.com/psf/black).

PR #123, opened 2024-08-08
Snapshot: Add possibility to have only one micro simulation object

tjwsch:
Adds functionality to only create one micro simulation object in the snapshot computation if all micro simulations are initialized with the same configuration and no information is transferred from one solve call to another.
Checklist:


tjwsch:
I added a comment regarding when it is useful to apply this option

PR #127, opened 2024-08-20
Fix snapshot status attribute and log completion only once

tjwsch:
The status attribute written to the database is inconsistent in the snapshot feature, this PR fixes that. Additionally, the completion of the snapshot computation is now only logged once instead of being logged by every MPI rank.

Checklist:

### DIFF
--- a/micro_manager/snapshot/snapshot.py
+++ b/micro_manager/snapshot/snapshot.py
@@ -129,7 +129,6 @@ class MicroManagerSnapshot(MicroManager):
             self._data_storage.write_crashed_snapshots(
                 self._output_file_path, self._crashed_snapshots
             )
-        self._data_storage.set_status(self._output_file_path, "none")
 
         # Merge output files
         if self._is_parallel:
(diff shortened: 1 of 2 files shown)


micro_manager/snapshot/snapshot.py
```diff
@@ -144,7 +143,10 @@ def solve(self) -> None:
                     list_of_output_files,
                     self._parameter_space_size,
                 )
-        self._logger.info("Snapshot computation completed.")
+        else:
+            self._data_storage.set_status(self._output_file_path, "finished")
+        if self._rank == 0:
+            self._logger.info("Snapshot computation completed.")
```
IshaanDesai, line 149: Is it possible that rank 0 finishes the snapshot computation before other ranks? If yes, there should be a MPI Wait command before the logging.

tjwsch: It shouldn't because rank 0 merges the output files from all ranks after completion in a parallel run. The completion statement is written after rank 0 merges the results.

PR #130, opened 2024-09-16
Put computation of counting active steps inside the adaptivity variant condition

### LINKED ISSUE #122: Overcounted active steps when using adaptivity in every implicit iteration
When using adaptivity in every implicit iteration, the count of active steps is still incremented at the beginning of every time window when writing a checkpoint in [micro_manager.py#L202](https://github.com/precice/micro-manager/blob/develop/micro_manager/micro_manager.py#L202). The increment should be guarded by the check for adaptivity based on time windows (line 169).

Note: The setting of `active_sim_ids` in between could also be moved to the conditional block. It is set separately for the case of adaptivity in every implicit iteration without being used in between.

IshaanDesai:
Fixes #122 

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Micro Manager changelog
 
+## latest
+
+- Put computation of counting active steps inside the adaptivity variant `if` condition https://github.com/precice/micro-manager/pull/130
+
 ## v0.5.0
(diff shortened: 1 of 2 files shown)


mathiskelm (approved):
Great, thanks for adding the fix.

PR #131, opened 2024-09-16
Add info about adaptivity tuning parameters

IshaanDesai:
In the configuration documentation, information about the adaptivity tuning parameters was missing. This PR adds information, which is taken from [this paper](https://doi.org/10.1016/j.amc.2020.125933).

Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -121,6 +121,12 @@ Parameter | Description
 `every_implicit_iteration` | If True, adaptivity is calculated in every implicit iteration. <br> If False, adaptivity is calculated once at the start of the time window and then reused in every implicit time iteration.
 `similarity_measure`| Similarity measure to be used for adaptivity. Can be either `L1`, `L2`, `L1rel` or `L2rel`. By default, `L1` is used. The `rel` variants calculate the respective relative norms. This parameter is *optional*.
 
+The primary tuning parameters for adaptivity are the history parameter $$ \Lambda $$, the coarsening constant $$ C_c $$, and the $$ C_r $$. Their effects can be interpreted as:
+
+- Higher values of the history parameter $$ \Lambda $$ imply lower significance of the adaptivity state in the previous timestep on the state in the current timestep.
+- Higher values of the coarsening constant $$ C_c $$ imply that more active simulations from the previous timestep will remain active in the current timestep.
+- Higher values of the refining constant $$ C_r $$ imply that less inactive points from the previous timestep will become active in the current timestep.
+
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -121,6 +121,12 @@ Parameter | Description
 `every_implicit_iteration` | If True, adaptivity is calculated in every implicit iteration. <br> If False, adaptivity is calculated once at the start of the time window and then reused in every implicit time iteration.
 `similarity_measure`| Similarity measure to be used for adaptivity. Can be either `L1`, `L2`, `L1rel` or `L2rel`. By default, `L1` is used. The `rel` variants calculate the respective relative norms. This parameter is *optional*.
 
+The primary tuning parameters for adaptivity are the history parameter $$ \Lambda $$, the coarsening constant $$ C_c $$, and the $$ C_r $$. Their effects can be interpreted as:
```
mathiskelm, line 124:
```suggestion
The primary tuning parameters for adaptivity are the history parameter $$ \Lambda $$, the coarsening constant $$ C_c $$, and the refining constant $$ C_r $$. Their effects can be interpreted as:
```

docs/configuration.md
```diff
@@ -121,6 +121,12 @@ Parameter | Description
...
+
+- Higher values of the history parameter $$ \Lambda $$ imply lower significance of the adaptivity state in the previous timestep on the state in the current timestep.
```
mathiskelm, line 126: I would stick to the similarity distance, rather than the adaptivity state, of previous time steps having an impact on the current measure and adaptivity decision.
```suggestion
- Higher values of the history parameter $$ \Lambda $$ imply lower significance of the similarity measures in the previous timestep on the similarity measure and thus adaptivity state in the current timestep.
```

docs/configuration.md
```diff
@@ -121,6 +121,12 @@ Parameter | Description
...
+- Higher values of the coarsening constant $$ C_c $$ imply that more active simulations from the previous timestep will remain active in the current timestep.
+- Higher values of the refining constant $$ C_r $$ imply that less inactive points from the previous timestep will become active in the current timestep.
```
mathiskelm, line 128: Might be good to point out that $$ C_r $$ affects both thresholds, but probably best kept simple.

IshaanDesai: I try to refrain from talking about thresholds. Would it be helpful to explain the logic of how the constants are used to determine thresholds which are then actually used to decide whether a simulation is active or inactive?

PR #133, opened 2024-11-07
Change logging strategy and add new logging for adaptivity metrics

IshaanDesai:
This PR adds the following:

- A new logging framework with a dedicated `Logger` class. In this class, a logger is generated and logging options are provided for logging via one one rank, any rank of choice, or all ranks. Different types of logging are possible. Currently the options are `info` and `error`.
- **Breaking change**: changes the configuration parameter `config_file_name` to `precice_config_file_name` to differentiate it from the Micro Manager JSON configuration file.
- Add optional parameter `output_cpu_time` to `adaptivity_settings` which will output CPU time of the adaptivity computation on every rank. The data is written to preCICE with the intention of being exported. All the micro simulations (coupling vertices) on one rank will have the same scalar value of the adaptivity computation CPU time.
- **Breaking change**: Change the name of the diagnostics output `micro_sim_time` to `solve_cpu_time`.

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## latest
 
+- Improve logging by wrapping Python logger in a class https://github.com/precice/micro-manager/pull/133
 - Add information about adaptivity tuning parameters https://github.com/precice/micro-manager/pull/131
 - Put computation of counting active steps inside the adaptivity variant `if` condition https://github.com/precice/micro-manager/pull/130
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -13,7 +13,7 @@ The Micro Manager is configured with a JSON file. An example configuration file 
 {
     "micro_file_name": "micro_solver",
     "coupling_params": {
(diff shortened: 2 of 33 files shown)


Fujikawas:
Would be good to put the "number of active simulations" into a format which is more machine-readable than the current text style.

uekerman:
Could https://github.com/precice/precice/issues/2015 be an alternative technical solution?

IshaanDesai:
> Could [precice/precice#2015](https://github.com/precice/precice/issues/2015) be an alternative technical solution?

Potentially yes. I have implemented something very close to this, with additionally handling logging in parallel. Would it be possible to tell preCICE to log somethings only on one rank, and other things on all ranks? Also, such logging API would need to be added to the Python bindings.

uekerman:
> Would it be possible to tell preCICE to log somethings only on one rank, and other things on all ranks?

Yes, this is part of the runtime configuration for the existing preCICE logging already now.

PR #143, opened 2025-01-24
Fix bug in handling of vector data

IshaanDesai:
The Micro Manager could not handle the case where a micro simulation returns vector data in the form of a list. This was because the data collection for adaptivity was done on a predefined `numpy.array` data structure. Now the data collection is done in a Python list and the list is converted to a NumPy array only just before the data is used for the similarity calculation.

Checklist:


mathiskelm (approved):
Fixes issue #144 also in my application. Looks good!

PR #164, opened 2025-05-09
Update the documentation on the website only when there is a release

IshaanDesai:
@uekerman @MakisH does this make sense? Right now, for every merge to develop, the website gets updated. In my opinion this is too frequent, and also misleading because oftentimes users work with the latest release and not with the latest develop state.

Checklist:

### DIFF
--- a/.github/workflows/update-website.yml
+++ b/.github/workflows/update-website.yml
@@ -2,7 +2,7 @@ name: Update website
 on:
   push:
     branches:
-      - 'develop'
+      - 'main'
     paths:
       - 'docs/**'
 jobs:


MakisH (commented):
For the micro-manager specifically, depends what you want.

For the rest of the components, this is pretty much how we did it before (rendering from the master branches), but we quickly ended up in tricky situations:

1. pressure to release, just to update the documentation
2. having to distinguish between bugfixes and the rest of the development in the documentation
3. forgetting to merge master to develop

I would advise against this, to keep things simpler for you. Alternative is using the version alerts:

https://precice.org/docs-meta-cheatsheet.html

IshaanDesai:
Thanks for the important points

> 
>     1. pressure to release, just to update the documentation

I have been updating the documentation lesser and lesser in the last few months, so I am thinking if such a pressure would really be the case.

>     2. having to distinguish between bugfixes and the rest of the development in the documentation

You mean bugfixes in the documentation itself? Or something else? I do not fully get your point here.

>     3. forgetting to merge master to develop

This step is part of the release guide and is never missed :)

uekerman:
> You mean bugfixes in the documentation itself? Or something else? I do not fully get your point here.

Exactly, you fix a typo in the documentation and the fix is only rendered once you release. Imagine a user doing this. Or you stumbling over the typo again. "Didn't I fix this before?". I agree with Makis' suggestion.

IshaanDesai:
> 
> Exactly, you fix a typo in the documentation and the fix is only rendered once you release. Imagine a user doing this. Or you stumbling over the typo again. "Didn't I fix this before?". I agree with Makis' suggestion.

Okay, this makes sense. I will add version information where relevant, but stick to building the website from the develop branch.

PR #206, opened 2025-12-12
Model Instancing to reduce memory footprint

Snapex2409:
Idea behind this is the following:

Currently, we save for each micro model its input, output as well as a full copy of the model itself.
IO data cannot easily be optimized, but as this should not be too much usually, certainly less than 1MB per model instance, this should not matter too much.

The critical issue is the full model itself. Depending on its implementation, the internal data structures might use up a significant amount of memory. In the case of certain ROMs this is easily >1MB, leading to very high memory consumption.

However, often the model is stateless. Therefore, the internal data structure do not need to be replicated for each micro-sim, rather can be shared.

Towards this, here is a first implementation, that maintains one "compute" instance for each sharable micro-sim type. 

Checklist:


Snapex2409:
This PR was split into multiple parts and can be closed.

PR #211, opened 2025-12-30
Fix URL in installation.md

MakisH:
In https://precice.org/tooling-micro-manager-installation.html

there is a link to https://precice.org/tooling-micro-manager-running.html/

which should be https://precice.org/tooling-micro-manager-running.html

I would even argue that these URLs should be absolute, not relative, since they can be clicked from this repository.

Checklist:

- I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`. -> N/A
- If necessary, I made changes to the documentation and/or added new content. -> N/A

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -15,7 +15,7 @@ The Micro Manager package on PyPI is [micro-manager-precice](https://pypi.org/pr
 pip install micro-manager-precice
 ```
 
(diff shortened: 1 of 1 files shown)


IshaanDesai (approved):
Thanks. I did not notice this.

PR #218, opened 2026-02-08
Patched Micro Simulation Interface

Snapex2409:
Changed loading of micro sim class:
- added wrapper class that handles some checks and acts as a "class"
- introduced abstraction layers to later support remote micro sim execution

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## latest
 
-- Add abstraction layers to micro simulations to support more features [#218](https://github.com/precice/micro-manager/pull/218)
+- Added abstraction layers to micro simulations to support more features [#218](https://github.com/precice/micro-manager/pull/218)
 
 ## v0.8.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## latest
 
-- Add abstraction layers to micro simulations to support more features [#217](https://github.com/precice/micro-manager/pull/217)
+- Add abstraction layers to micro simulations to support more features [#218](https://github.com/precice/micro-manager/pull/218)
 
 ## v0.8.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Micro Manager changelog
 
+## latest
+
+- Add abstraction layers to micro simulations to support more features [#217](https://github.com/precice/micro-manager/pull/217)
+
 ## v0.8.0
 
 - Conformed to naming standard in precice/tutorials [#215](https://github.com/precice/micro-manager/pull/215)
--- a/micro_manager/adaptivity/model_adaptivity.py
+++ b/micro_manager/adaptivity/model_adaptivity.py
@@ -4,7 +4,11 @@ Class ModelAdaptivity provides methods to change micro simulation resolution on 
 from typing import Union, Optional
 
 from micro_manager.config import Config
-from micro_manager.micro_simulation import create_simulation_class, load_backend_class, MicroSimulationClass
+from micro_manager.micro_simulation import (
+    create_simulation_class,
+    load_backend_class,
+    MicroSimulationClass,
+)
 from micro_manager.tools.logging_wrapper import Logger
 from micro_manager.tools.misc import clamp_in_range
 
@@ -283,7 +287,9 @@ class ModelAdaptivity:
         """
         return len(self._model_classes)
 
-    def get_resolution_sim_class(self, resolution: Union[int, np.ndarray]) -> Union[MicroSimulationClass, list[MicroSimulationClass]]:
+    def get_resolution_sim_class(
+        self, resolution: Union[int, np.ndarray]
+    ) -> Union[MicroSimulationClass, list[MicroSimulationClass]]:
         """
         Looks up the class associated with the provided resolution.
 
--- a/micro_manager/adaptivity/model_adaptivity.py
+++ b/micro_manager/adaptivity/model_adaptivity.py
@@ -3,8 +3,8 @@ Class ModelAdaptivity provides methods to change micro simulation resolution on 
 """
 from typing import Union, Optional
 
-from ..config import Config
-from ..micro_simulation import create_simulation_class, load_backend_class
+from micro_manager.config import Config
+from micro_manager.micro_simulation import create_simulation_class, load_backend_class, MicroSimulationClass
 from micro_manager.tools.logging_wrapper import Logger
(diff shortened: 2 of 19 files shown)


micro_manager/adaptivity/model_adaptivity.py
```diff
@@ -150,19 +151,86 @@ def switch_models(
             if cur_res[idx] == tgt_res[idx]:
                 continue
 
-            sim_state = sims[idx].get_state()
-            sim_id = sims[idx].get_global_id()
-            sims[idx] = self.get_resolution_sim_class(tgt_res[idx])(sim_id)
-            sims[idx].set_state(sim_state)
+            sim = sims[idx]
+            gid = sim.get_global_id()
+            tgt_cls = self.get_resolution_sim_class(tgt_res[idx])
```
IshaanDesai: I assume this is target class. Please use the full form to clearly state the variable.

micro_manager/adaptivity/model_adaptivity.py
```diff
@@ -150,19 +151,86 @@ def switch_models(
...
+            new_state_exists = key_new in sim.attachments
+            sim.attachments[key] = sim.get_state()
+
+            sim_new = self.tgt_cls(gid, late_init=new_state_exists)
+            sim_new.attachments = sim.attachments
+            sim_new.attachments[key_new] = sim_new.get_state()
+
+            if new_state_exists:
+                sim_new_state = sim.attachments[key_new]
+                sim_new.set_state(sim_new_state)
```
IshaanDesai, line 182: Can you add comments to improve code readability?

micro_manager/adaptivity/model_adaptivity.py
```diff
@@ -150,19 +151,86 @@ def switch_models(
...
+
+            sims[idx] = sim_new
+
+    def update_states(
+        self,
+        sims: list,
+        active_sim_ids: Optional = None,
+    ):
+        """
+        Stores the current state of the current model into local buffers.
```
IshaanDesai:
```suggestion
        Updates the current state of the current model in the local buffers.
```
To synchronize with the name of the function.

micro_manager/adaptivity/model_adaptivity.py
```diff
@@ -150,19 +151,86 @@ def switch_models(
...
+
+        for idx in range(size):
+            if not active_sims[idx]:
+                continue
+
+            sim = sims[idx]
+            key = f"{sim.name}-state"
+            sim.attachments[key] = sim.get_state()
+
+    def write_back_states(
```
IshaanDesai, line 212: Why not name this load_states?

micro_manager/adaptivity/model_adaptivity.py
```diff
@@ -150,19 +151,86 @@ def switch_models(
...
+            sim.set_state(sim.attachments[key])
 
     def check_convergence(
         self,
         locations: np.ndarray,
         t: float,
-        inputs: list[dict],
-        prev_output: Optional[dict],
+        inputs: list,
+        prev_output: Optional,
```
IshaanDesai: Is there a particular reason you remove the sub-type hints from everywhere? Does this have something to do with the superMUC compatibility branch?

micro_manager/micro_simulation.py
IshaanDesai, line 1: Just for my understanding, what is the reason to have several class hierarchy levels here?

micro_manager/micro_simulation.py
```diff
@@ -4,8 +4,202 @@ class MicroSimulation. A global ID member variable is defined for the class Simu
...
+    @attachments.setter
+    def attachments(self, value):
+        self._external_data = value
+
+    @property
+    def name(self):
+        return self._name
+
+
+class MicroSimulationClassAdapter:
```
IshaanDesai: Let us find a different name here. If we say *Adapter*, in the preCICE world, it immediately means an adapter for a physics-based solver.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Micro Manager changelog
 
+## latest
+
+- Add abstraction layers to micro simulations to support more features [#218](https://github.com/precice/micro-manager/pull/218)
```
IshaanDesai:
```suggestion
- Added abstraction layers to micro simulations to support more features [#218](https://github.com/precice/micro-manager/pull/218)
```
Changelog entries are written in past tense.

IshaanDesai (commented):
I understand almost all the changes, and they make a lot of sense! Thank you! I have a few questions and mostly naming-style suggestions. Once these are addressed, we can merge this PR.

micro_manager/adaptivity/model_adaptivity.py
Snapex2409: The thought process here was: we only "update" our global multi-state buffer instead of "loading" all and replacing the previous one.

micro_manager/adaptivity/model_adaptivity.py
Snapex2409: Yes, at some point I had merged with your SuperMUC branch. By doing so those were removed.

However, when using FenicsX this becomes obsolete again as we require a newer Python version that now supports these subtypes...

micro_manager/micro_simulation.py
Snapex2409: The ClassAdapter exists to be compatible with the rest, so we do not need to rewrite everything too much. It was needed to perform comparisons between classes more elegantly.

The Hierarchy of SimWrapper using SimLocal(SimInterface):
This is used such that we can later add a SimRemote class. The Wrapper can then delegate to the correct implementation. This could've been done using inheritance. But the main issue was: I wanted a location to store the multi-states cleanly. The SimLocal / SimRemote objects should not require those nor know about those.

micro_manager/micro_simulation.py
Snapex2409: Ok... I just used Adapter as that is the common design pattern name

In theory MicroSimulationClass should work, or?

micro_manager/adaptivity/model_adaptivity.py
IshaanDesai: > However, when using FenicsX this becomes obsolete again as we require a newer Python version that now supports these subtypes...

In that case, lets introduce these subtypes again.

micro_manager/micro_simulation.py
IshaanDesai: Got it, makes sense!

micro_manager/micro_simulation.py
IshaanDesai: `MicroSimulationClass` should work yes. Lets use that name.

IshaanDesai (approved):

PR #219, opened 2026-02-19
Implented Tasking Model

Snapex2409:
Tasking Model was implemented here to support MPI based Micro-Sim solve computation.
Workers are required to create an isolate MPI process.

Checklist:


micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+import psutil
+import socket
+import struct
+import subprocess
+import os
+from abc import ABC, abstractmethod
+from typing import Any, Dict, Optional
+from mpi4py import MPI
+
+# task import workaround
```
IshaanDesai: Can you make this comment more descriptive?

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+        self.inter_comm = None
+
+    @classmethod
+    def create_workers(
+        cls, worker_exec: str, mpi_args: Optional, n_workers: int
+    ) -> "MPIConnection":
+        comm = MPI.COMM_SELF
+        conn = cls()
+        conn.inter_comm = comm.Spawn(
+            f"python {worker_exec}",
```
IshaanDesai: Why `python` here? Maybe `Micro Manager ` workers make more sense?

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+            args=mpi_args or [],
+            maxprocs=n_workers,
+        )
+        return conn
+
+    @classmethod
+    def connect_to_micromanager(cls, parent_comm) -> "MPIConnection":
```
IshaanDesai:
```suggestion
    def connect_to_micro_manager(cls, parent_comm) -> "MPIConnection":
```

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+        conn.inter_comm = parent_comm
+        return conn
+
+    def send(self, dst_id: int, obj: Any) -> None:
+        data = pickle.dumps(obj, protocol=pickle.HIGHEST_PROTOCOL)
+        self.inter_comm.send(data, dest=dst_id, tag=0)
+
+    def recv(self, src_id: int) -> Any:
+        data = self.inter_comm.recv(source=src_id, tag=1)
+        return pickle.loads(data)
```
IshaanDesai, line 111: I understand what these functions are doing, but can you add short doctstrings so that in the future it is clear what these functions are doing? Not just in terms of their functionality, but also in the context of task-based implementation.

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+        cls, worker_exec: str, launcher: list, host: str, n_workers: int, env_opts: dict
+    ) -> "SocketConnection":
+        # create listening socket with ephemeral port
+        server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        server.bind((host, 0))  # kernel picks free port
+        server.listen()
+        port = server.getsockname()[1]
+
+        executable = [
+            "python",
```
IshaanDesai, line 136:
```suggestion
            "Micro Manager",
```
perhaps?

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+        conn = cls()
+        for wid in range(n_workers):
+            sock, _ = server.accept()
+            conn.sockets[wid] = sock
+
+        server.close()
+        return conn
+
+    @classmethod
+    def connect_to_micromanager(
```
IshaanDesai:
```suggestion
    def connect_to_micro_manager(
```

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+        if not header:
+            raise EOFError
+        (size,) = struct.unpack("!Q", header)
+        payload = b""
+        while len(payload) < size:
+            chunk = sock.recv(size - len(payload))
+            if not chunk:
+                raise EOFError
+            payload += chunk
+        return pickle.loads(payload)
```
IshaanDesai, line 212: Similar comment to what is above: please add docstrings.

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+        MPI implementation
+
+    Returns
+    -------
+    list
+        pinning args
+    """
+    args = []
+    rank = MPI.COMM_WORLD.Get_rank()
+    size = MPI.COMM_WORLD.Get_size()
```
IshaanDesai, line 236: Above in line 41, `MPI.COMM_SELF` is used. Do we need to be consistent here?

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+
+    options = {}
+    if mpi_impl == "intel":
+        if "I_MPI_INFO_LCPU" in os.environ:
+            nCPU = int(os.environ["I_MPI_INFO_LCPU"]) / 2
+        else:
+            raise RuntimeError("Cannot Determine CPU count")
+
+        if os.path.exists("./hosts.micro"):
+            with open("./hosts.micro", "r") as f:
```
IshaanDesai: I assume the file name `hosts.micro` is hard-coded because that is how it is named in the scripts I have. Can we parse this name? Perhaps via the configuration?

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+                hosts = f.readlines()
+        else:
+            raise RuntimeError("Cannot Nodes")
+
+        if size % len(hosts) != 0:
+            raise RuntimeError("Num ranks must be divisible by number of hosts")
```
IshaanDesai:
```suggestion
            raise RuntimeError("Number of ranks must be divisible by number of hosts")
```

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+
+        mm_ppn = size // len(hosts)
+        node_idx = rank // mm_ppn
+        node = hosts[node_idx].replace("\n", "")
+
+        locations_int = list(os.sched_getaffinity(0))
+        if any([pos >= nCPU for pos in locations_int]):
+            raise RuntimeError("Pinning Error: trying to pin to HT CPU")
```
IshaanDesai: Just for my understanding, what does *HT* refer to?

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+
+        options.update(
+            {
+                "I_MPI_DEBUG": "5",
+                "I_MPI_PIN": "1",
+                "I_MPI_PIN_CELL": "core",
+                "I_MPI_PIN_DOMAIN": "1",
+                "I_MPI_PIN_PROCESSOR_LIST": locations,
+            }
+        )
```
IshaanDesai, line 266: Can you write a short comment on what these options are? Even a link to the documentation would do.

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+
+    candidates = []
+
+    # Iterate over preferred interfaces first
+    if preferred_ifaces:
+        for name in preferred_ifaces:
+            if name not in addrs:
+                continue
+            for a in addrs[name]:
+                if a.family == socket.AF_INET and not a.address.startswith("127."):
```
IshaanDesai, line 308: Do IPs starting with `127.` have a classification? Can you mention that here in a comment?

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+                    return a.address
+
+    # Fallback: iterate all interfaces
+    for name, iface_addrs in addrs.items():
+        for a in iface_addrs:
+            if a.family == socket.AF_INET:
+                ip = a.address
+                if not ip.startswith("127.") and not ip.startswith("169.254."):
```
IshaanDesai, line 317: Similar comment as above.

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+    ----------
+    worker_exec : str
+        path to worker executable
+    n_workers : int
+        number of worker processes, must be > 1 otherwise returns None
+    backend : str
+        mpi or socket
+    is_slurm : bool
+        is our system slurm based?
+
```
IshaanDesai, line 352: `mpi_impl` input parameter is missing from docstring.

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+    conn = None
+
+    # MPI BACKEND (non-Slurm only)
+    if backend == "mpi":
+        if is_slurm:
+            raise RuntimeError(
+                "MPI backend is not supported under Slurm. "
+                "Use socket backend instead."
+            )
+        comm = MPI.COMM_WORLD
```
IshaanDesai, line 369: Does this conflict with the use of `MPI.COMM_SELF` in the class `MPIConnection`?

micro_manager/tasking/connection.py
```diff
@@ -0,0 +1,339 @@
...
+            host=host,
+            n_workers=n_workers,
+            env_opts=pin_options,
+        )
+
+    from ..micro_simulation import load_backend_class
+
+    for worker_id in range(n_workers):
+        conn.send(worker_id, RegisterAllTask(load_backend_class))
+        conn.recv(worker_id)
```
IshaanDesai, line 417: A short comment would help here.

micro_manager/tasking/task.py
IshaanDesai, line 1: The child classes of the `Task` class need docstrings. I see that they are essentially doing what the MicroSimulation class does, but collectively. Is my understanding correct here?

micro_manager/tasking/worker_main.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    args = parser.parse_args()
+
+    rank = MPI.COMM_WORLD.Get_rank()
+    size = MPI.COMM_WORLD.Get_size()
+    worker_id = rank
+
+    conn, dst_id, src_id = None, 0, 0
+    if args.backend == "mpi":
+        print(f"Launched Worker with MPI rank: {rank}")
```
IshaanDesai: It would be better to parse these statements via the logger. Also below.

micro_manager/config.py
```diff
@@ -182,6 +188,27 @@ def _read_json(self, config_file_name):
...
+                    self._task_num_workers = self._data["tasking"]["num_workers"]
+                if self._task_is_slurm and backend == "mpi":
+                    raise Exception("MPI backend not supported on SLURM systems.")
+                if "mpi_impl" in self._data["tasking"]:
+                    self._task_mpi_impl = self._data["tasking"]["mpi_impl"]
+                    if self._task_mpi_impl not in ["open", "intel"]:
+                        raise Exception("mpi_impl must be either 'open' or 'intel'.")
+        except BaseException:
+            self._logger.log_info_rank_zero(
+                "No or incorrect tasking information provided. Micro manager will compute locally."
```
IshaanDesai: The second sentence needs to be more descriptive.

micro_manager/micro_simulation.py
```diff
@@ -8,6 +8,8 @@ class MicroSimulation. A global ID member variable is defined for the class Simu
 import inspect
 import importlib as ipl
 
+from .tasking.task import *
```
IshaanDesai: Please stick to explicit importing, so not `*`.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## latest
 
+- Added support to run micro simulation in separate process with workers [#219](https://github.com/precice/micro-manager/pull/219)
```
IshaanDesai:
```suggestion
- Added support to run micro simulations in separate processes with workers [#219](https://github.com/precice/micro-manager/pull/219)
```

IshaanDesai (changes requested):
I did a first pass over the implementation, very nice work 👏 
I understood most of the things. My comments are mostly about documentation, and some minor things I noticed. One thing we should consider is whether and how we can unit-test the tasking mechanism. Some testing would be valuable here, even if it is with 1 task and 2 workers, or something bare-bones.

micro_manager/tasking/connection.py
Snapex2409: No, that would not work directly. For that to work we would need the micro_manager package to export the worker_main.py to global PATH when building the pip/conda package.

Currently workers are not further MM instances, but rather just the smallest possible runnable py-programs that are able to run MicroSims.
Therefore, I need to call python here.

micro_manager/tasking/connection.py
Snapex2409: same as above, won't work

micro_manager/tasking/connection.py
IshaanDesai: Okay, got it.

micro_manager/tasking/connection.py
Snapex2409: No, these are 2 different things. When spawning we use COMM_SELF such that only an inter communicator between the parent MM rank to its workers gets created.
If we used COMM_WORLD all MM ranks would get connected to all workers, which is not what we want.

micro_manager/tasking/connection.py
Snapex2409: That is a hyper thread. But I removed that check now as it was taylored for the SuperMUC CPU layout and not needed in general.

micro_manager/tasking/task.py
Snapex2409: Yes, that's right. They tell the worker what to call next. One could replace this whole Task hierarchy using a pure text based protocol, but the current solution is probably more modern/easier to extend.

micro_manager/tasking/connection.py
IshaanDesai: Got it :+1:

IshaanDesai (approved):

PR #220, opened 2026-02-23
Fix duplicate micro simulations for macro-points on rank boundaries #102

### LINKED ISSUE #102: Duplicate micro-simulations on interior boundaries of macro-domain decomposition
Problem:
When using macro-domain decomposition each MPI rank is given a bounding box, which is passed to preCICE to obtain coordinates of coupled points contained therein. If such a coupled macro-point lies on the boundary between two bounding boxes it will be included in both and the Micro Manager currently creates a micro simulation for both duplicates of the macro-point and passes their output to preCICE.
- preCICE interprets this as two contributions to the participant's solution and accumulates them, producing incorrect results.
- duplicate micro simulations receive identical macro-data and with no/global adaptivity identical output, wasting computational effort
- with local adaptivity the two micro-simulations may have different associated active simulations and produce different outputs (differences should be minor if similarity measure is not overly generous)

Suggestion:
The Micro Manager should eliminate duplicate micro simulations to save computational time and avoid incorrect data being received by the macro simulation.

AdityaGupta716:
Closes #102
Problem:
When running the Micro Manager in parallel, each MPI rank is assigned a bounding box. Macro-points lying exactly on the boundary between two ranks were being included in both, causing the Micro Manager to create duplicate micro simulations for the same point. This led to:

Incorrect results in preCICE, which accumulates both outputs
Wasted computation from running identical micro simulations
Potentially different outputs with local adaptivity

Fix:
In get_local_sims_and_macro_coords, added an is_upper_boundary check per axis. A rank now uses <= for its upper bound only if that bound coincides with the global domain boundary — otherwise < is used. This ensures every boundary point is assigned to exactly one rank.
Changes:

micro_manager/domain_decomposition.py: Fixed boundary condition logic in get_local_sims_and_macro_coords
tests/unit/test_domain_decomposition.py: Added test_no_duplicate_coords_on_boundary_2d to verify no point is double-counted across ranks

### DIFF
--- a/micro_manager/domain_decomposition.py
+++ b/micro_manager/domain_decomposition.py
@@ -121,17 +121,17 @@ class DomainDecomposer:
 
         micro_sims_on_rank = 0
         for position in macro_coords:
-    inside = True
-    for d in range(len(ranks_per_axis)):
-        is_upper_boundary = np.isclose(local_mesh_bounds[d * 2 + 1], macro_bounds[d * 2 + 1])
-        if not (
-            position[d] >= local_mesh_bounds[d * 2]
-            and (position[d] <= local_mesh_bounds[d * 2 + 1] if is_upper_boundary else position[d] < local_mesh_bounds[d * 2 + 1])
-        ):
-            inside = False
-            break
-    if inside:
-        macro_coords_on_this_rank.append(position)
-        micro_sims_on_rank += 1
+            inside = True
+            for d in range(len(ranks_per_axis)):
+                is_upper_boundary = np.isclose(local_mesh_bounds[d * 2 + 1], macro_bounds[d * 2 + 1])
+                if not (
+                    position[d] >= local_mesh_bounds[d * 2]
+                    and (position[d] <= local_mesh_bounds[d * 2 + 1] if is_upper_boundary else position[d] < local_mesh_bounds[d * 2 + 1])
+                ):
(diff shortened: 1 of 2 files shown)


AdityaGupta716:
@MakisH @IshaanDesai Plz review

IshaanDesai:
Good effort, but the implementation does not address the issue as stated. The Micro Manager decomposes the domain based on the macro-scale domain bounds provided by the user via the JSON configuration file. The bounds of the partitioned domains are then handed to preCICE via

https://github.com/precice/micro-manager/blob/5dc8f4f9f3e9f120c19a151b3a01d3eca619ff88/micro_manager/micro_manager.py#L465-L467

preCICE then returns vertex coordinates in each of the paritioned domain via:

https://github.com/precice/micro-manager/blob/5dc8f4f9f3e9f120c19a151b3a01d3eca619ff88/micro_manager/micro_manager.py#L478-L481

It is here that the variable `mesh_vertex_coords` could contain duplicate vertices. By duplicate, I mean two MPI ranks could have the same vertex coordinate in this NumPy array. To fix the issue, the vertex coordinate information needs to be gathered on one rank, duplicates filtered, and then each rank given its own unique set of vertices.

I will close this PR for now. If an implementation to fix the issue is done, please reopen the PR or open a new PR.

PR #221, opened 2026-02-23
Add dependency checks with clear error messages on missing packages #70

### LINKED ISSUE #70: Add check if dependencies are installed correctly when Micro Manager is installed via pip
As suggested in https://github.com/openjournals/joss-reviews/issues/5842#issuecomment-1781849246 there needs to be a check to ensure that the dependencies are installed correctly when the Micro Manager is installed via pip.

AdityaGupta716:
Closes #70
Problem:
When the Micro Manager is installed via pip and a required dependency is missing, users receive a cryptic error deep in the import chain rather than a clear message explaining what is missing and how to fix it.
Fix:
Added a _check_dependencies() function in micro_manager/__init__.py that runs at startup and checks for all required packages (pyprecice, numpy, mpi4py, psutil). If any are missing, a single clear ImportError is raised listing the missing packages and the exact pip install command to resolve the issue.
Changes:

micro_manager/__init__.py: Added _check_dependencies() function called at module load time
tests/unit/test_dependency_check.py: Added tests that directly call _check_dependencies() to verify a clear error is raised for missing dependencies and no error when all are present


AdityaGupta716:
@MakisH @IshaanDesai Plz review

micro_manager/__init__.py
```diff
@@ -1,6 +1,29 @@
...
 import os
 
+
+def _check_dependencies():
+    missing = []
+    for package, import_name in [
+        ("pyprecice", "precice"),
+        ("numpy", "numpy"),
+        ("mpi4py", "mpi4py"),
+        ("psutil", "psutil"),
```
IshaanDesai: It would be good to check the versions too, and not just the dependencies.

micro_manager/__init__.py
```diff
@@ -13,6 +36,7 @@
 
 
 def main():
+    _check_dependencies()
```
IshaanDesai: The main intention is to check if everything is correctly installed by doing something like `micro-manager-precice --test`. If we check the dependencies in the `main` function, it means that if a dependency is installed incorrectly, the Micro Manager will explode when running. This is not ideal from a user perspective.

tests/unit/test_dependency_check.py
IshaanDesai, line 1: Rather than adding unit tests, it would be preferable to add a GitHub Actions workflow that tests the pip installation of the Micro Manager, and then runs the dependency check by doing something like `micro-manager-precice --test` or `micro-manager-precice --test-dependencies`.

IshaanDesai (changes requested):
Good effort, but missing the mark on what the issue requires. Please take a look at my comments.

AdityaGupta716:
Hi @IshaanDesai, addressed all three points — added version checking, moved the check to --test-dependencies flag, and replaced unit tests with a GitHub Actions workflow. Also added packaging to pyproject.toml.
One doubt — should _check_dependencies() also be exposed as a public API so that users importing micro_manager programmatically (not via CLI) can also call it directly to verify their setup?

IshaanDesai:
> One doubt — should _check_dependencies() also be exposed as a public API so that users importing micro_manager programmatically (not via CLI) can also call it directly to verify their setup?

Not necessary to expose this via the API. If a user installs the Micro Manager, I would expect them to be able to run the dependencies check via the executable.

Good work so far! Please address the failing CI.

AdityaGupta716:
Hi @IshaanDesai, fixed both CI failures switched to the precice/precice:nightly container and fixed a subtle bug where heavy imports were crashing before the dependency check could even run. All checks passing now!

.github/workflows/test-pip-install.yml
```diff
@@ -0,0 +1,51 @@
+name: Test pip installation and dependency check
+
+on:
+  push:
+    branches:
+      - develop
+      - p2
+  pull_request:
+    branches:
+      - develop
```
IshaanDesai:
```suggestion
      - develop
      - main
  pull_request:
    branches:
      - "*"
```

.github/workflows/test-pip-install.yml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
+  test-pip-install:
+    runs-on: ubuntu-latest
+    container: precice/precice:nightly
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
IshaanDesai:
```suggestion
        uses: actions/checkout@v6
```

micro_manager/__init__.py
```diff
@@ -1,27 +1,86 @@
...
 
-    is_snapshot_possible = True
-except ImportError:
-    is_snapshot_possible = False
+    required = {
+        "pyprecice": ("precice", "3.2"),
+        "numpy": ("numpy", None),
+        "mpi4py": ("mpi4py", None),
+        "psutil": ("psutil", None),
+    }
```
IshaanDesai: Can this information be read from the `pyproject.toml` file? It is duplicated here.

micro_manager/__init__.py
```diff
@@ -31,10 +90,25 @@ def main():
...
     if not os.path.isabs(config_file_path):
         config_file_path = os.getcwd() + "/" + config_file_path
 
+    try:
+        from .snapshot.snapshot import MicroManagerSnapshot
+
+        is_snapshot_possible = True
+    except ImportError:
+        is_snapshot_possible = False
+
```
IshaanDesai: Why include this here now?

IshaanDesai (commented):
Thank you for the effort here. We are going in the right direction. I have some comments. Once addressed, this can be ready for merging.

micro_manager/__init__.py
AdityaGupta716: sry ,That's a duplicate removing it

micro_manager/__init__.py
AdityaGupta716: switched to importlib.metadata.requires("micro-manager-precice") which reads the deps from the installed package metadata directly, so no duplication. also filters out optional extras like snapshot and sklearn.

micro_manager/__init__.py
```diff
@@ -1,27 +1,95 @@
...
-def main():
+    missing = []
+    version_errors = []
+
+    for package, (import_name, min_version) in required.items():
+        try:
+            __import__(import_name)
+            if min_version:
+                installed_version = importlib.metadata.version(package)
+                from packaging.version import Version
```
IshaanDesai: Please move this line to be immediately after line 8.

IshaanDesai (commented):
Almost there. Just one last thing I noticed. Also, please add a changelog entry.

micro_manager/__init__.py
AdityaGupta716: done

IshaanDesai (approved):
Thanks a lot for getting this done!

PR #223, opened 2026-02-24
Add TestPyPI publishing workflow for release PRs #108

### LINKED ISSUE #108: Need CI to test package publishing workflow
The [package publishing workflow](https://github.com/precice/micro-manager/blob/develop/.github/workflows/pythonpublish.yml) is run every time a release is made. If changes to the workflow or package setup are done, the way to test the workflow is by doing a release candidate. This is a cumbersome way to test the publishing workflow. It would be better to have a workflow which publishes to [TestPyPI](https://test.pypi.org/) every time a release PR is opened, and only when it passes, the release is allowed.

AdityaGupta716:
Closes #108

Adds a new workflow that publishes to TestPyPI every time a PR targeting main is opened or updated.

Changes:
- Added `publish-to-testpypi.yml` workflow that builds the package and publishes to TestPyPI on release PRs
- Uses trusted publishing (id-token: write) same as the existing PyPI workflow
- Only triggers on PRs to main, keeping it specific to release PRs

Note: A TestPyPI trusted publisher needs to be configured by a maintainer at test.pypi.org for this workflow to work.


AdityaGupta716:
Hey @MakisH @IshaanDesai, plz review! TestPyPI trusted publisher needs to be configured at test.pypi.org by a maintainer for the workflow to actually run

IshaanDesai (commented):
Thanks for the contribution. The intention of publishing on TestPyPI after releasing is to ensure that the release to PyPI is successful. Keeping this in mind, can you please check whether we can configure the `release.yml` workflow to run only after `publish-to-testpypi.yml` succeeds?

Also, please take a look at the [release.yml](https://github.com/precice/micro-manager/blob/develop/.github/workflows/release.yml) workflow and use the latest versions of the actions.

AdityaGupta716:
@IshaanDesai hi, addressed both comments, Restructured release.yml so PyPI publish only runs after publish-to-testpypi succeeds via needs, and updated all action versions to match the rest of the repo.

PR #224, opened 2026-02-24
 Expose MicroSimulationInterface as public abstract base class for user subclassing #65

### LINKED ISSUE #65: Provide a micro simulation abstract class which users can inherit/create a subclass
In https://github.com/openjournals/joss-reviews/issues/5842#issuecomment-1758196669, @mbkumar suggested that the Micro Manager could provide an abstract class for the micro simulation code, which the users could subclass or inherit, and the give the name of their class to the Micro Manager via the JSON config file. This is a cleaner way to ensure that the users stick to the desired API while converting their micro simulations into callable libraries.

AdityaGupta716:
Closes #65

Exposes MicroSimulationInterface as a public abstract base class that users can inherit from when writing their micro simulations. This ensures users stick to the required API and get clear errors if they miss a method.

Changes:
- Added full docstrings with examples to all abstract methods in MicroSimulationInterface
- Exported MicroSimulationInterface from __init__.py so users can do `from micro_manager import MicroSimulationInterface`
- Updated load_backend_class() to warn users with a DeprecationWarning if their class does not inherit from MicroSimulationInterface (will become a hard error in a future version)
- Updated examples/python-dummy/micro_dummy.py to demonstrate the new inheritance pattern

### DIFF
--- a/micro_manager/micro_simulation.py
+++ b/micro_manager/micro_simulation.py
@@ -211,11 +211,12 @@ class MicroSimulationLocal(MicroSimulationInterface):
 
 
 class MicroSimulationRemote(MicroSimulationInterface):
-    def __init__(self, gid, late_init, num_ranks, conn, cls_path):
+    def __init__(self, gid, late_init, num_ranks, conn, cls_path, sim_cls):
         self._cls_path = cls_path
         self._gid = gid
         self._num_ranks = num_ranks
         self._conn = conn
+        self._sim_cls = sim_cls
 
         construct_cls = ConstructLateTask if late_init else ConstructTask
         for worker_id in range(self._num_ranks):
@@ -291,6 +292,12 @@ class MicroSimulationRemote(MicroSimulationInterface):
 
         return result
 
+    def requires_initialize(self) -> bool:
+        return self._sim_cls.initialize is not MicroSimulationInterface.initialize
+
+    def requires_output(self) -> bool:
+        return self._sim_cls.output is not MicroSimulationInterface.output
+
 
 class MicroSimulationWrapper(MicroSimulationInterface):
     """
@@ -303,7 +310,7 @@ class MicroSimulationWrapper(MicroSimulationInterface):
 
         if num_ranks > 1 and conn is not None:
             self._impl = MicroSimulationRemote(
-                global_id, late_init, num_ranks, conn, cls_path
+                global_id, late_init, num_ranks, conn, cls_path, sim_cls
             )
         else:
             self._impl = MicroSimulationLocal(global_id, late_init, sim_cls)
--- a/micro_manager/micro_simulation.py
+++ b/micro_manager/micro_simulation.py
@@ -396,7 +396,8 @@ class MicroSimulationClass:
         # If the class inherits from MicroSimulationInterface, use requires_initialize()
         try:
             if issubclass(self._sim_cls, MicroSimulationInterface):
-                if not test_instance.requires_initialize():
+                # Check at class level without relying on instance state
+                if self._sim_cls.initialize is MicroSimulationInterface.initialize:
                     return False, False
         except TypeError:
             pass  # pybind11 classes may not support issubclass
@@ -462,11 +463,8 @@ class MicroSimulationClass:
         # If the class inherits from MicroSimulationInterface, use requires_output()
         try:
             if issubclass(self._sim_cls, MicroSimulationInterface):
-                instance = self._sim_cls.__new__(self._sim_cls)
-                try:
-                    return instance.requires_output()
-                except Exception:
-                    pass
+                # Check at class level without instantiating
+                return self._sim_cls.output is not MicroSimulationInterface.output
         except TypeError:
             pass  # pybind11 classes may not support issubclass
 
--- a/micro_manager/micro_simulation.py
+++ b/micro_manager/micro_simulation.py
@@ -376,10 +376,9 @@ class MicroSimulationClass:
         """
         Check whether the micro simulation class implements ``initialize``.
 
-        If the class inherits from ``MicroSimulationInterface``, use
-        ``requires_initialize()`` to determine whether the method is overridden.
-        Otherwise, fall back to ``hasattr`` for backwards compatibility with
-        classes that do not inherit from the interface (e.g. pybind11 classes).
+        Since ``load_backend_class`` guarantees that ``self._sim_cls`` always
+        inherits from ``MicroSimulationInterface``, we can rely on
+        ``requires_initialize()`` directly. No ``issubclass`` guard is needed.
 
         Parameters
         ----------
@@ -393,20 +392,7 @@ class MicroSimulationClass:
         tuple[bool, bool]
             (has_initialize, requires_initial_data)
         """
-        # If the class inherits from MicroSimulationInterface, use requires_initialize()
-        try:
-            if issubclass(self._sim_cls, MicroSimulationInterface):
-                # Check at class level without relying on instance state
-                if self._sim_cls.initialize is MicroSimulationInterface.initialize:
-                    return False, False
-        except TypeError:
-            pass  # pybind11 classes may not support issubclass
-
-        has_init = hasattr(self._sim_cls, "initialize")
-        if not has_init:
-            return False, False
-        callable_init = callable(getattr(self._sim_cls, "initialize"))
-        if not callable_init:
+        if not test_instance.requires_initialize():
             return False, False
 
         has_args = False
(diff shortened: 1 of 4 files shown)


AdityaGupta716:
Hey @MakisH @IshaanDesai, plz review! Let me know if the deprecation warning approach works for you or if you'd prefer a hard error right away.

micro_manager/micro_simulation.py
```diff
@@ -20,32 +20,124 @@ class MicroSimulation. A global ID member variable is defined for the class Simu
...
+    def initialize(self, *args, **kwargs):
+        """
+        Initialize the micro simulation. Called once before the coupling loop starts.
+
+        Parameters
+        ----------
+        initial_data : dict, optional
+            Initial data passed from the Micro Manager.
+        """
+        pass
```
Snapex2409: In the current process we check whether initialize (and also output) are members of the simulation class. If so, then we call them accordingly. By enforcing their existence in the interface, we will likely run into issues during that.

Perhaps, add further abstractmethods like `def requires_initialize(self) -> bool: ...` and `requires_output`.
Just to be compatible with the rest of the code base.

micro_manager/micro_simulation.py
Snapex2409, line 1: Please add typing to all args (as much as possible at least), we are trying to transition to fully typed variables.

micro_manager/micro_simulation.py
```diff
@@ -20,32 +20,124 @@ class MicroSimulation. A global ID member variable is defined for the class Simu
...
+            Global ID to assign.
+        """
         pass
 
     @abstractmethod
     def output(self):
+        """
+        Optional output method called after each solve step.
+        """
         pass
```
Snapex2409, line 150: same as above with initialize

micro_manager/micro_simulation.py
```diff
@@ -302,7 +394,18 @@ def check_output(self):
...
+    if not issubclass(cls, MicroSimulationInterface):
+        import warnings
+        warnings.warn(
+            "The MicroSimulation class in '{}' does not inherit from MicroSimulationInterface. "
+            "Please update your class definition to: "
+            "class MicroSimulation(MicroSimulationInterface). "
+            "In a future version this will become an error.".format(path_to_micro_file),
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
Snapex2409: This might become an issue for us. If we only used python micro simulation files that should be fine. However, we also use pybind11 modules.

At least based on a quick research pybind11 does not exactly allow for inheritance when exporting a module. (check in the examples for the cpp dummy, whether it works with this)

Snapex2409 (changes requested):
Overall great idea and would be nice to have. Some minor changes requested, and please check regarding C++ bindings. Thank You.

Snapex2409:
PS: run `pre-commit run -a -v` for example to get rid of the formatting issue in the CI.

AdityaGupta716:
@Snapex2409  hi , addressed all the review comments, Removed @abstractmethod from initialize and output and made them optional with default pass implementations, added requires_initialize() and requires_output() for class-level override detection, and updated check_initialize() and check_output() to use these instead of hasattr. Also wrapped the issubclass check in a try/except for  pybind11 compatibility,, added type hints throughout, and ran black to fix the pre-commit failure.
Couldn't test the C++ dummy directly since it needs pybind11 compiled locally, but the logic should handle it fine since non-inheriting classes fall back to the existing hasattr path.

micro_manager/micro_simulation.py
```diff
@@ -291,18 +447,54 @@ def check_initialize(self, test_instance, test_input):
...
 
 def load_backend_class(path_to_micro_file):
     CLS_NAME = "MicroSimulation"
-    return getattr(ipl.import_module(path_to_micro_file, CLS_NAME), CLS_NAME)
+    cls = getattr(ipl.import_module(path_to_micro_file, CLS_NAME), CLS_NAME)
+    try:
+        inherits = issubclass(cls, MicroSimulationInterface)
+    except TypeError:
+        # pybind11 classes may not support issubclass checks
+        inherits = False
```
Snapex2409, line 585: Lets make this a uniform interface for any provided micro simulation, such that the output of `load_backend_class` always adheres to the `MicroSimulationInterface`. Right now this might get a bit wobbly when a pybind11 or other object is provided.

If `issubclass` fails, we could dynamically create a class that inherits from the interface, then delegates all method calls to the loaded class. If the loaded class has initialize or output implementation, then we explicitly add an override in the dynamically created class (also delagates, but will then make `requires_output` return true)
The general delegation can probably be done via:
```Python
    def __getattr__(self, name):
        return getattr(self._instance, name)
```

Dynamic Class creation will probably need to be something in the lines of:
```Python
    sim_class_body = """
def __init__(self, global_id):
    micro_simulation_class.__init__(self, global_id)
    self._global_id = global_id

def get_global_id(self) -> int:
    return self._global_id
    """
    sim_class_dict = {}
    local_globals = {
        "__builtins__": __builtins__,
        "micro_simulation_class": micro_simulation_class,
    }
    exec(sim_class_body, local_globals, sim_class_dict)
    # print(sim_class_dict)
    sim_class = type(sim_class_name, (micro_simulation_class,), sim_class_dict)

    return sim_class

```
However, in the proposed variant the sim_class_body will need to be altered depending on whether output or initialize exist.

micro_manager/micro_simulation.py
```diff
@@ -291,18 +447,54 @@ def check_initialize(self, test_instance, test_input):
...
+        Returns
+        -------
+        bool
+            True if the micro simulation class has a callable ``output`` method.
+        """
+        # If the class inherits from MicroSimulationInterface, use requires_output()
+        try:
+            if issubclass(self._sim_cls, MicroSimulationInterface):
+                # Check at class level without instantiating
+                return self._sim_cls.output is not MicroSimulationInterface.output
```
Snapex2409: By enforcing MicroSimulationInterface (see comment regarding `load_backend_class`) we can now call `requires_output` here and drop the issubclass check.

micro_manager/micro_simulation.py
```diff
@@ -245,7 +370,38 @@ def __call__(self, gid, *, late_init=False):
...
+            Sample input data used to probe whether ``initialize`` accepts arguments.
+
+        Returns
+        -------
+        tuple[bool, bool]
+            (has_initialize, requires_initial_data)
+        """
+        # If the class inherits from MicroSimulationInterface, use requires_initialize()
+        try:
+            if issubclass(self._sim_cls, MicroSimulationInterface):
```
Snapex2409: Same as `check_output`:
By enforcing MicroSimulationInterface (see comment regarding `load_backend_class`) we can now call `requires_initialize` here and drop the issubclass check.

Snapex2409 (changes requested):
Overall thank you for your efforts. Nicely done. Please consider what I wrote in the comments. Let me know if I should clarify further.

AdityaGupta716:
@Snapex2409   hi done with the changes ,  Wrapped non-conforming classes in a dynamic adapter in load_backend_class so it always returns a proper MicroSimulationInterface subclass. The adapter delegates everything via __getattr__ and only adds initialize/output overrides if the original class actually has them — so requires_initialize()/requires_output() work correctly for all cases including pybind11. check_initialize and check_output are now clean with no guards.
Also caught a bug while implementing — the wrapper chain (MicroSimulationWrapper, MicroSimulationLocal, MicroSimulationRemote) was unconditionally overriding initialize and output, which made requires_initialize()/requires_output() always return True. Fixed that too.
Still can't test the C++ dummy locally, but the logic should handle it correctly since the TypeError from issubclass triggers the wrap path and that test fail seems unrelated

micro_manager/micro_simulation.py
```diff
@@ -245,12 +389,29 @@ def __call__(self, gid, *, late_init=False):
...
         return self._sim_cls
 
-    def check_initialize(self, test_instance, test_input):
-        has_init = hasattr(self._sim_cls, "initialize")
-        if not has_init:
-            return False, False
-        callable_init = callable(getattr(self._sim_cls, "initialize"))
-        if not callable_init:
+    def check_initialize(
+        self, test_instance: object, test_input: dict
```
Snapex2409:
```suggestion
        self, test_instance: MicroSimulationInterface, test_input: dict
```

micro_manager/micro_simulation.py
```diff
@@ -20,34 +20,159 @@ class MicroSimulation. A global ID member variable is defined for the class Simu
...
+        Parameters
+        ----------
+        micro_sim_input : dict
+            Input data from the macro simulation.
+        dt : float
+            Time step size.
+
+        Returns
+        -------
+        dict
```
Snapex2409:
```suggestion
        micro_sim_output : dict
```

micro_manager/micro_simulation.py
```diff
@@ -20,34 +20,159 @@ class MicroSimulation. A global ID member variable is defined for the class Simu
...
 
     @abstractmethod
-    def set_state(self, state):
+    def get_state(self) -> object:
+        """
+        Return the current state of the micro simulation for checkpointing.
+
+        Returns
+        -------
+        object
```
Snapex2409:
```suggestion
        state : object
```

micro_manager/micro_simulation.py
```diff
@@ -20,34 +20,159 @@ class MicroSimulation. A global ID member variable is defined for the class Simu
...
 
     @abstractmethod
-    def set_global_id(self, global_id):
+    def get_global_id(self) -> int:
+        """
+        Return the global ID of this micro simulation instance.
+
+        Returns
+        -------
+        int
```
Snapex2409:
```suggestion
        global_id : int
```

micro_manager/micro_simulation.py
```diff
@@ -20,34 +20,159 @@ class MicroSimulation. A global ID member variable is defined for the class Simu
...
         pass
 
+    def requires_initialize(self) -> bool:
+        """
+        Return True if this simulation class overrides the ``initialize`` method.
+        The Micro Manager calls this to determine whether initialization is needed.
+
+        Returns
+        -------
+        bool
```
Snapex2409:
```suggestion
        requires_initialize : bool
```

micro_manager/micro_simulation.py
```diff
@@ -20,34 +20,159 @@ class MicroSimulation. A global ID member variable is defined for the class Simu
...
+        return type(self).initialize is not MicroSimulationInterface.initialize
+
+    def requires_output(self) -> bool:
+        """
+        Return True if this simulation class overrides the ``output`` method.
+        The Micro Manager calls this to determine whether output is needed.
+
+        Returns
+        -------
+        bool
```
Snapex2409:
```suggestion
        requires_output : bool
```

micro_manager/micro_simulation.py
```diff
@@ -245,12 +389,29 @@ def __call__(self, gid, *, late_init=False):
...
+        Parameters
+        ----------
+        test_instance : object
+            An instance of the micro simulation class used for signature probing.
+        test_input : dict
+            Sample input data used to probe whether ``initialize`` accepts arguments.
+
+        Returns
+        -------
+        tuple[bool, bool]
```
Snapex2409:
```suggestion
        check_result : tuple[bool, bool]
```

micro_manager/micro_simulation.py
```diff
@@ -289,20 +450,147 @@ def check_initialize(self, test_instance, test_input):
...
+        """
+        Check whether the micro simulation class implements ``output``.
+
+        Since ``load_backend_class`` guarantees that ``self._sim_cls`` always
+        inherits from ``MicroSimulationInterface``, we can rely on
+        ``requires_output()`` directly at the class level.
+
+        Returns
+        -------
+        bool
```
Snapex2409:
```suggestion
        check_result : bool
```

micro_manager/micro_simulation.py
```diff
@@ -289,20 +450,147 @@ def check_initialize(self, test_instance, test_input):
...
+        class_body += """
+def output(self):
+    return self._wrapped.output()
+"""
+
+    class_dict = {}
+    exec(class_body, {"wrapped_cls": cls, "__builtins__": __builtins__}, class_dict)
+
+    wrapper_cls = type(
+        "MicroSimulationWrapper_{}".format(cls.__name__),
```
Snapex2409:
```suggestion
        "CompatibilityWrapper_{}".format(cls.__name__),
```

Snapex2409: This would otherwise always be MicroSimulationWrapper_MicroSimulation

Snapex2409 (commented):
Looks very good. Thank you 👍 

Could you determine why the test fails?

The suggested changes are just to have the same docstring style as in other files.

AdityaGupta716:
@Snapex2409 Hi , applied all the docstring suggestions 
Looked into the CI failure too — reproduced it on upstream/develop directly with no changes from this PR, so it's pre-existing. The root cause is in test_set_state itself: it calls del sim mid-test which triggers cleanup on the MicroSimulationRemote object, then immediately tries to reuse the same worker connection to send a GetStateTask. The workers have already lost state at that point, so recv() hits an EOFError. It's a resource management issue in the test unrelated to our changes.
Happy to fix it in this PR if that works for you

Snapex2409:
@AdityaGupta716  I think you did not check out develop while testing, but that's ok. Issue is resolved now.
Thank you for your help.

Apart from that: As you already mentioned, that is not clean resource management. But I am already addressing that in another PR. So, no need to deal with that here.

AdityaGupta716:
@Snapex2409   i Ran test_set_state on upstream/develop locally  it gets stuck in setUp at spawn_local_workers, hanging on server.accept() waiting for a worker to connect. prterun can't find the python executable so the worker never spawns and the socket just hangs. Had to kill it manually 

Screenshot 2026-03-04 at 3 12 04 AM

Snapex2409:
@AdityaGupta716 Looks like your mpi is not able to find to find python. What python environment are you using venv, conda or other? Is it perhaps not in PATH?

Snapex2409 (approved):

AdityaGupta716:
@Snapex2409 ahh got it my bad was using venv and the path is python3 not python , got it

PR #225, opened 2026-02-24
Add CI workflow to check test coverage

### LINKED ISSUE #55: Add CI to check test coverage
It would be good to check test coverage for every change via a pull request. As the Micro Manager is written in Python, any solution would involve using [Coverage.py](https://coverage.readthedocs.io/en/7.2.7/). There are ways to automate this in GitHub Actions via existing actions or Codecov, but it seems easier to implement a simple Python script which computes the test coverage and checks if it is above or below a predefined threshold.

AdityaGupta716:
Closes #55

Adds a workflow to check test coverage on every push and pull request using Coverage.py.

Changes:
- Added check_coverage.py script that runs coverage report and fails if coverage is below a predefined threshold (currently set to 70%)
- Added check-coverage.yml workflow that runs all unit tests with coverage and calls the script to check the threshold

The threshold can be adjusted in check_coverage.py as the codebase grows.


AdityaGupta716:
@MakisH  @IshaanDesai, plz review! The threshold is currently set to 70% in check_coverage.py .if u want smthing different let me know.

IshaanDesai:
@AdityaGupta716 can you please take a look at the CI output of the test coverage workflow? It seems like the workflow got stuck.

AdityaGupta716:
@IshaanDesai  hi, fixed the issue  the workflow was hanging because unittest discover was picking up the parallel tests (test_adaptivity_parallel.py, test_global_adaptivity_lb.py) which need MPI to run. changed it to run each test file explicitly instead — same approach as the existing run-unit-tests.yml. Also current coverage is around 22%, so the 70% threshold might need adjusting , happy to change it to whatever you think fits!

IshaanDesai:
Good work so far! I am surprised that the test coverage is not 70%, though I perceive it as having a certain value. The [CI log](https://github.com/precice/micro-manager/actions/runs/22776339810/job/66070558786?pr=225), however, shows 0% test coverage, which seems definitely incorrect. Can you please take a look at what is going on?

AdityaGupta716:
hi @IshaanDesai  fixed the 0% coverage issue! The problem was the .coverage file wasn't being found correctly. The workflow now runs all serial and parallel unit tests with mpirun and combines the results, giving a real coverage of ~61%. I've set the threshold to 60% for now plz review and let me know if you think it should be different

IshaanDesai:
@Snapex2409, to increase test coverage, some tests on things that you recently implemented have been added. Can you please go over them to ensure that they make sense?

micro_manager/interpolation.py
```diff
@@ -46,7 +46,7 @@ def get_nearest_neighbor_indices(
         return neighbor_indices
 
     def interpolate(self, neighbors: np.ndarray, point: np.ndarray, values):
-        """
+        r"""
```
Snapex2409, line 49:
```suggestion
        """
```

tests/unit/test_micro_simulation.py
```diff
@@ -0,0 +1,257 @@
...
+    def test_default_initialize_returns_none(self):
+        sim = MinimalSim(0)
+        self.assertIsNone(sim.initialize())
+
+    def test_default_output_returns_none(self):
+        sim = MinimalSim(0)
+        self.assertIsNone(sim.output())
+
+
+class TestMicroSimulationLocal(unittest.TestCase):
```
Snapex2409, line 75: If we test `MicroSimulationLocal` here, then we should do the same for `Remote`.

.github/workflows/check-coverage.yml
```diff
@@ -0,0 +1,65 @@
...
+          pip install .[sklearn,snapshot]
+          pip uninstall -y pyprecice
+
+      - name: Run serial unit tests with coverage
+        working-directory: micro-manager/tests/unit
+        env:
+          PYTHONPATH: .
+        run: |
+          . ../../.venv/bin/activate
+          python3 -m coverage run --parallel-mode --source=micro_manager -m unittest test_micro_manager
```
Snapex2409, line 77: Please add the `tasking` unit tests here too.
For those tests to pass we require OpenMPI 5. You can load and install a cached version. See `run-unit-tests.yml`. (just need the load and configure steps)

Snapex2409 (changes requested):
Overall, nicely done. With the requested changes, coverage should also increase a bit.

tests/unit/test_micro_simulation.py
AdityaGupta716: Done,Added TestMicroSimulationRemote with tests covering solve, get/set_state, get/set_global_id, initialize, output, requires_initialize, requires_output, destroy, and late_init  same as the Local tests.

.github/workflows/check-coverage.yml
AdityaGupta716: Added the OpenMPI 5 load/build/cache/configure steps from run-unit-tests.yml, and added test_tasking as a separate step with OMPI_ALLOW_RUN_AS_ROOT env vars set so it runs correctly in the container.

tests/unit/test_micro_simulation.py
Snapex2409: Looks good to me. Is there a reason why `TestMicroSimulationRemote` is at the end of the file and not directly after the local tests? If not, then could you just move it up.

Other than that, good work. Thank you for the tests. Helps a lot.

Snapex2409 (changes requested):
Just the small refactory, then we should be ready to merge in my opinion.
And good thinking to mock the connection here 👍

AdityaGupta716:
@Snapex2409 nah no reasons, moved it up

PR #227, opened 2026-02-25
Add /dev/shm space check before shared memory allocation in global adaptivity #217

### LINKED ISSUE #217: Global adaptivity is limited to shared memory on device
For documentation purposes mainly. I'm not sure to what extent this is a problem, but on some system shared memory is only 50% of RAM, which means that the user can only use 50% compared to local adaptivty option or no adaptivity.

```bash
It appears as if there is not enough space for /dev/shm/osc_sm.abc.1cbb0001.0.4 (the shared-memory backing
file). It is likely that your MPI job will now either abort or experience
performance degradation.

  Local host:  abc
  Space Requested: 70350139792 B
  Space Available: 66749136896 B
--------------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/USADR/ac135212/spack/var/spack/environments/examples_final/.spack-env/view/bin/micro-manager-precice", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/USADR/ac135212/spack/var/spack/environments/examples_final/.spack-env/._view/j2i6wdbluvyqkt5a3by5mzvsm7wvc5xw/lib/python3.12/site-packages/micro_manager/__init__.py", line 47, in main
    manager.initialize()
  File "/home/USADR/ac135212/spack/var/spack/environments/examples_final/.spack-env/._view/j2i6wdbluvyqkt5a3by5mzvsm7wvc5xw/lib/python3.12/site-packages/micro_manager/micro_manager.py", line 597, in initialize
    GlobalAdaptivityCalculator(
  File "/home/USADR/ac135212/spack/var/spack/environments/examples_final/.spack-env/._view/j2i6wdbluvyqkt5a3by5mzvsm7wvc5xw/lib/python3.12/site-packages/micro_manager/adaptivity/global_adaptivity.py", line 84, in __init__
    win = MPI.Win.Allocate_shared(nbytes, itemsize, comm=self._comm_node)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/mpi4py/MPI.src/Win.pyx", line 126, in mpi4py.MPI.Win.Allocate_shared
mpi4py.MPI.Exception: MPI_ERR_INTERN: internal error`
```

AdityaGupta716:
Closes #217 
Problem:
On some systems /dev/shm is limited to 50% of total RAM. When global adaptivity is used with a large number of simulations, the N×N similarity distance matrix allocated via MPI.Win.Allocate_shared can exceed this limit, causing a cryptic MPI internal error with no useful information for the user.
Fix:
Before calling MPI.Win.Allocate_shared, rank 0 now checks the available space in /dev/shm using shutil.disk_usage. If there is not enough space, a clear RuntimeError is raised telling the user exactly how much space is required vs available, and suggests either increasing /dev/shm size or switching to local adaptivity.
Note:
This is a documentation/UX fix — no logic or algorithm changes. The error was always there, it just surfaced as an unhelpful MPI crash. This makes it actionable for the user.


AdityaGupta716:
@IshaanDesai please review

IshaanDesai (commented):
Thanks for tackling this issue and proposing a solution. I will test this soon on a cluster where the memory-bound is expected to be hit and see if the correct exception pops up.

PR #229, opened 2026-02-25
Add CI workflow to check PyPI deployment after release #56

### LINKED ISSUE #56: Check package deployed on PyPI via CI
Currently there is no check if the package deployed to PyPI actually works. Such a check should be done once a package is deployed after a release. Running just the solver dummy would be sufficient.

AdityaGupta716:
Closes #56

Adds a workflow that triggers automatically after the "Upload Python Package" workflow completes successfully. It installs micro-manager-precice directly from PyPI and runs the Python dummy to verify the deployed package works correctly.


AdityaGupta716:
@MakisH @IshaanDesai Please review

PR #230, opened 2026-02-27
Fix duplicate micro simulations for macro-points on rank boundaries

### LINKED ISSUE #102: Duplicate micro-simulations on interior boundaries of macro-domain decomposition

AdityaGupta716:
Closes #102
Problem: When a macro-point lies exactly on the boundary between two rank bounding boxes, preCICE returns it to both ranks. This causes duplicate micro simulations, incorrect accumulated results in preCICE, and wasted computation.
Fix: After get_mesh_vertex_ids_and_coordinates(), each rank uses allgather to see all coords across all ranks, then drops any coord already owned by a lower-ranked rank — ensuring every vertex is processed by exactly one rank while preserving the preCICE ID-coord pairing safely.
Changes:

micro_manager/micro_manager.py: Added deduplication after get_mesh_vertex_ids_and_coordinates()
tests/unit/test_domain_decomposition.py: Removed incorrect test that tested wrong location

### DIFF
--- a/micro_manager/domain_decomposition.py
+++ b/micro_manager/domain_decomposition.py
@@ -135,56 +135,50 @@ class DomainDecomposer:
 
         return micro_sims_on_rank, macro_coords_on_this_rank
 
+    def filter_duplicate_coords(
+        self,
+        all_coords: list,
+        all_ids: list,
+    ) -> tuple[np.ndarray, np.ndarray]:
+        """
+        Filter out vertex coordinates that are already owned by a lower-ranked rank.
 
-def filter_duplicate_coords(
-    my_rank: int,
-    size: int,
-    all_coords: list,
-    all_ids: list,
-) -> tuple[np.ndarray, np.ndarray]:
-    """
-    Filter out vertex coordinates that are already owned by a lower-ranked rank.
+        When a macro-point lies exactly on the boundary between two rank bounding
+        boxes, preCICE returns it to both ranks. This function ensures every vertex
+        is processed by exactly one rank — the lowest-ranked rank that received it —
+        while preserving the preCICE ID-coord pairing.
 
-    When a macro-point lies exactly on the boundary between two rank bounding
-    boxes, preCICE returns it to both ranks. This function ensures every vertex
-    is processed by exactly one rank — the lowest-ranked rank that received it —
-    while preserving the preCICE ID-coord pairing.
+        Parameters
+        ----------
+        all_coords : list
+            List of numpy arrays, one per rank, containing vertex coordinates.
+        all_ids : list
+            List of arrays, one per rank, containing preCICE vertex IDs.
 
-    Parameters
-    ----------
-    my_rank : int
-        MPI rank of the calling process.
-    size : int
-        Total number of MPI processes.
-    all_coords : list
-        List of numpy arrays, one per rank, containing vertex coordinates.
-    all_ids : list
-        List of arrays, one per rank, containing preCICE vertex IDs.
+        Returns
+        -------
+        filtered_coords : numpy.ndarray
+            Vertex coordinates with duplicates removed.
+        filtered_ids : numpy.ndarray
+            preCICE vertex IDs corresponding to the filtered coordinates.
+        """
+        mesh_vertex_coords = np.array(all_coords[self._rank])
+        mesh_vertex_ids = np.array(all_ids[self._rank])
 
-    Returns
-    -------
-    filtered_coords : numpy.ndarray
-        Vertex coordinates with duplicates removed.
-    filtered_ids : numpy.ndarray
-        preCICE vertex IDs corresponding to the filtered coordinates.
-    """
-    mesh_vertex_coords = np.array(all_coords[my_rank])
-    mesh_vertex_ids = np.array(all_ids[my_rank])
+        seen_coords = set()
+        keep_mask = np.ones(len(mesh_vertex_coords), dtype=bool)
 
-    seen_coords = set()
-    keep_mask = np.ones(len(mesh_vertex_coords), dtype=bool)
-
-    for rank in range(size):
-        for i, coord in enumerate(all_coords[rank]):
-            coord_key = tuple(np.round(coord, decimals=10))
-            if rank < my_rank:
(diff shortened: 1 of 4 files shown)


AdityaGupta716:
Hi @IshaanDesai, opening a new PR as suggested. I used allgather instead of gather-scatter because the IDs returned by preCICE are local to each rank and used directly for reading/writing data — redistributing them risked breaking the ID-coord pairing. With allgather, each rank sees all coords and drops any already claimed by a lower-ranked rank, eliminating duplicates safely. Plz review

micro_manager/domain_decomposition.py
IshaanDesai, line 1: Changes in this file do not seem necessary. Please revert if you also think so.

micro_manager/micro_manager.py
```diff
@@ -480,6 +480,33 @@ def initialize(self) -> None:
...
+            # find globally duplicated coords, and remove them from this rank
+            # while preserving the preCICE ID-coord pairing.
+            all_coords = self._comm.allgather(self._mesh_vertex_coords)
+            all_ids = self._comm.allgather(self._mesh_vertex_ids)
+
+            # Build a global set of (coord, id) pairs seen by previous ranks
+            seen_coords = set()
+            keep_mask = np.ones(len(self._mesh_vertex_coords), dtype=bool)
+
+            for r in range(self._size):
```
IshaanDesai:
```suggestion
            for rank in range(self._size):
```

micro_manager/micro_manager.py
```diff
@@ -480,6 +480,33 @@ def initialize(self) -> None:
...
+                for i, coord in enumerate(all_coords[r]):
+                    coord_key = tuple(np.round(coord, decimals=10))
+                    if r < self._rank:
```
IshaanDesai:
```suggestion
                    if rank < self._rank:
```

micro_manager/micro_manager.py
```diff
@@ -480,6 +480,33 @@ def initialize(self) -> None:
...
+                        # Mark coords already claimed by earlier ranks
+                        seen_coords.add(coord_key)
+                    elif r == self._rank:
```
IshaanDesai:
```suggestion
                    elif rank == self._rank:
```

tests/unit/test_domain_decomposition.py
IshaanDesai, line 1: Changes in this file are just formatting, please revert.

IshaanDesai (commented):
Thanks for the contribution. Please take a look at my comments. Additionally, could you please add a test to ensure that duplicate coordinate filtering is done as expected?

AdityaGupta716:
@IshaanDesai  hi, reverted the changes in domain_decomposition.py as allgather does those fixes at higher level and the formatting-only changes in test_domain_decomposition.py, renamed r to rank in the deduplication loop, and added a new TestDuplicateCoordFiltering test class with three cases — no duplicates, a shared boundary coord between two ranks, and a shared coord across three ranks.

IshaanDesai:
@AdityaGupta716, as a general practice: once you address a review comment by making the relevant changes, you can hit *Resolve conversation*. You don't have to wait for the developer to do this, unless the developer explicitly states so. This makes it easier for the developer, me in this case, to see the changes that are yet to be addressed.

tests/unit/test_domain_decomposition.py
```diff
@@ -95,3 +95,94 @@ def test_rank7_out_of_16_3d(self):
...
         self.assertTrue(np.allclose(mesh_bounds, [0.75, 1, -2, 0, -2, 8]))
+
+
+class TestDuplicateCoordFiltering(TestCase):
+    """
+    Test that duplicate vertex coordinates returned by preCICE on rank boundaries
+    are correctly deduplicated, with lower-ranked ranks taking ownership.
+    """
+
+    def _run_dedup(self, my_rank, size, all_coords, all_ids):
```
IshaanDesai: The test should run the actual implementation from `micro_manager.py`, not a copy of it. Take a look at the other tests in this file to see how the functionality can be tested in isolation.

tests/unit/test_domain_decomposition.py
```diff
@@ -95,3 +95,94 @@ def test_rank7_out_of_16_3d(self):
         )
 
         self.assertTrue(np.allclose(mesh_bounds, [0.75, 1, -2, 0, -2, 8]))
+
+
+class TestDuplicateCoordFiltering(TestCase):
+    """
+    Test that duplicate vertex coordinates returned by preCICE on rank boundaries
+    are correctly deduplicated, with lower-ranked ranks taking ownership.
```
IshaanDesai:
```suggestion
    are correctly filtered, with lower-ranked ranks taking ownership.
```

IshaanDesai (changes requested):
Good work so far! The implementation looks good, and so do the tests, but the tests should test the actual implementation and not a copy of it. Please take a look at my comments.

AdityaGupta716:
@IshaanDesai Hi done with the changes plz check it out !

IshaanDesai (commented):
We are almost there! Can you please make the function `filter_duplicate_coords` a member function of the `DomainDecomposer` class? That would simplify the importing and also future maintenance.

AdityaGupta716:
hi @IshaanDesai done with the changes!

IshaanDesai (approved):
Great work! Thanks for going through multiple review cycles.

PR #238, opened 2026-03-31
Fix lazy initialization

IshaanDesai:
Fixes to lazy initialization when ...

- ... there are ranks without an active problem.
- ... there are ranks with no micro problem to initialize.
- ... lazy initialization is used without adaptivity.

Checklist:


micro_manager/micro_manager.py
```diff
@@ -811,6 +808,38 @@ def initialize(self) -> None:
...
+                    self._data_for_adaptivity[name][i] = initial_micro_data_list[i][
+                        name
+                    ]
+
+        self._micro_sims_have_output = False
+        if hasattr(micro_problem_cls, "output") and callable(
+            getattr(micro_problem_cls, "output")
+        ):
+            self._micro_sims_have_output = True
+
```
Snapex2409: This is already being handled in line 797

micro_manager/micro_manager.py
```diff
@@ -811,6 +808,38 @@ def initialize(self) -> None:
             self._rank,
         )
 
+        # If lazy initialization is on, initial states of inactive simulations need to be determined
+        if self._lazy_init:
+            # Prepare data structure for collective communication
+            if initial_micro_data:
+                initial_micro_data_list = [
+                    dict(zip(initial_micro_data, t))
```
Snapex2409, line 816: Not 100% sure, but could it be that you need here: `initial_micro_data.values()`

Snapex2409 (approved):
Overall makes sense to me.

micro_manager/micro_manager.py
IshaanDesai: I think the code is correct. `zip(initial_micro_data)` unpacks the keys of the dictionary, which I want to put together with the tuples of values unpacked from `zip(*initial_micro_data.values())`

PR #239, opened 2026-04-01
Reduce size of similarity distance matrix for memory efficiency

IshaanDesai:
It is not necessary to store the N² matrix of similarity distances. Only storing the lower (or upper) triangular part is sufficient. This PR implements such storage and access. This is done to mitigate the memory costs of global adaptivity.

Checklist:

### DIFF
--- a/micro_manager/adaptivity/adaptivity.py
+++ b/micro_manager/adaptivity/adaptivity.py
@@ -58,15 +58,17 @@ class AdaptivityCalculator:
 
         self._max_similarity_dist = 0.0
 
+        self._nsims = nsims
+
         # is_sim_active: 1D array having state (active or inactive) of each micro simulation
         # Start adaptivity calculation with all sims active
         # This array is modified in place via the function update_active_sims and update_inactive_sims
-        self._is_sim_active = np.array([True] * nsims, dtype=np.bool_)
+        self._is_sim_active = np.array([True] * self._nsims, dtype=np.bool_)
 
         # sim_is_associated_to: 1D array with values of associated simulations of inactive simulations. Active simulations have None
         # Active sims do not have an associated sim
         # This array is modified in place via the function associate_inactive_to_active
-        self._sim_is_associated_to = np.full((nsims), -2, dtype=np.intc)
+        self._sim_is_associated_to = np.full((self._nsims), -2, dtype=np.intc)
 
         self._just_deactivated: list[int] = []
 
@@ -131,7 +133,9 @@ class AdaptivityCalculator:
                 # The axis is later reduced with a norm.
                 data_vals = np.expand_dims(data_vals, axis=1)
 
-            self._similarity_dists += dt * self._similarity_measure(data_vals)
+            self._similarity_measure(data_vals)
+
+        self._similarity_dists *= dt
 
     def _associate_inactive_to_active(self) -> None:
         """
@@ -149,10 +153,11 @@ class AdaptivityCalculator:
             # Begin with a large distance to trigger the search for the most similar active sim
             dist_min = dist_min_start_value
             for active_id in active_ids:
+                flat_idx = self._map2dto1d([inactive_id, active_id])
                 # Find most similar active sim for every inactive sim
-                if self._similarity_dists[inactive_id, active_id] < dist_min:
+                if self._similarity_dists[flat_idx] < dist_min:
                     associated_active_id = active_id
-                    dist_min = self._similarity_dists[inactive_id, active_id]
+                    dist_min = self._similarity_dists[flat_idx]
 
             self._sim_is_associated_to[inactive_id] = associated_active_id
 
@@ -171,7 +176,10 @@ class AdaptivityCalculator:
         tag : bool
             True if the inactive simulation needs to be activated, False otherwise.
         """
-        dists = self._similarity_dists[inactive_id, active_ids]
+        flat_ids = [
+            self._map2dto1d([inactive_id, active_id]) for active_id in active_ids
+        ]
+        dists = self._similarity_dists[flat_ids]
 
         # If inactive sim is not similar to any active sim, activate it
         return min(dists) > self._ref_tol
@@ -194,14 +202,15 @@ class AdaptivityCalculator:
         """
         for active_id_2 in active_ids:
             if active_id != active_id_2:  # don't compare active sim to itself
+                flat_idx = self._map2dto1d([active_id, active_id_2])
                 # If active sim is similar to another active sim, deactivate it
-                if self._similarity_dists[active_id, active_id_2] < self._coarse_tol:
+                if self._similarity_dists[flat_idx] < self._coarse_tol:
                     return True
         return False
 
     def _get_similarity_measure(
         self, similarity_measure: str
-    ) -> Callable[[np.ndarray], np.ndarray]:
+    ) -> Callable[[np.ndarray], None]:
         """
         Get similarity measure to be used for similarity calculation
 
@@ -228,7 +237,7 @@ class AdaptivityCalculator:
                 'Similarity measure not supported. Currently supported similarity measures are "L1", "L2", "L1rel", "L2rel".'
             )
 
-    def _l1(self, data: np.ndarray) -> np.ndarray:
+    def _l1(self, data: np.ndarray) -> None:
         """
         Calculate L1 norm of data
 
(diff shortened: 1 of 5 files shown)


micro_manager/adaptivity/adaptivity.py
```diff
@@ -295,19 +299,59 @@ def _l2rel(self, data: np.ndarray) -> np.ndarray:
...
+                flat_idx = self._map2dto1d([i, j])
+                self._similarity_dists[flat_idx] += np.linalg.norm(rel_diff, ord=2) * dt
+
+    def _map1dto2d(self, flat_idx: int) -> tuple:
+        """
+        Map a 1D index to 2D coordinates in strictly lower triangular part.
+
+        Parameters
+        ----------
+        flat_idx : int
```
Snapex2409: No loop is needed if we were in a strictly upper triangular storage.
To get rid of the diagonal we can emulate the same effect by just considering a storage matrix of size n-1. In there the (non strict) upper triangular part has our desired dimensions.

With this all the following operations can be vectorized using numpy operations.

micro_manager/adaptivity/adaptivity.py
```diff
@@ -295,19 +299,59 @@ def _l2rel(self, data: np.ndarray) -> np.ndarray:
...
+            1D index to be mapped
 
         Returns
         -------
-        similarity_dists : numpy array
-            Updated 2D array having similarity distances between each micro simulation pair
+        coord : tuple
+            2D coordinates (i, j) corresponding to the 1D index, where i > j
```
Snapex2409: Right, I forgot about that. We need to enforce either i < j or i > j.
Using the idea of strict upper tri we will have i < j.

micro_manager/adaptivity/adaptivity.py
```diff
@@ -295,19 +299,59 @@ def _l2rel(self, data: np.ndarray) -> np.ndarray:
...
-        # Add small epsilon to avoid division by zero (invalid value warning) when both are 0
-        eps = np.finfo(np.float64).eps
-        relative = pointwise_diff / np.maximum(denom, eps)
-        return np.linalg.norm(relative, ord=2, axis=-1)
+        # Find row i: solve i(i-1)/2 <= flat_idx < i(i+1)/2
+        i = 1
+        while i * (i + 1) // 2 <= flat_idx:
+            i += 1
+        j = flat_idx - i * (i - 1) // 2
+        return (i, j)
```
Snapex2409: should also allow for list of inputs
flat_idx can then be int or list[int]
instead of this we can now do:
```suggestion
        if type(flat_idx) == int:
            flat_idx = [flat_idx]
        flat_idx = np.array(flat_idx)
        
        i = flat_idx // (self._nsims - 1)
        j = (flat_idx % (self._nsims - 1)) + 1 # shift by one to transfrom back from n-1 mat to n mat
        return np.vstack((i, j), dtype=np.int32).T
```

micro_manager/adaptivity/adaptivity.py
```diff
@@ -295,19 +299,59 @@ def _l2rel(self, data: np.ndarray) -> np.ndarray:
...
+        """
+        i, j = coords[0], coords[1]
+        if i <= j:
+            # For symmetric matrix access, swap if needed
+            i, j = j, i
+        if i <= j:
+            raise ValueError(
+                f"Invalid indices for lower triangular: ({i}, {j}). Need i > j."
+            )
+        return i * (i - 1) // 2 + j
```
Snapex2409: similar here:
will assume a list of coords as input [[0, 1], [0, 2], ...]

```suggestion
        coords = np.array(coords, dtype=np.int32).reshape(-1, 2)
        # assert i < j invariant
        coords = np.sort(coords, axis=-1)
        i, j = coords[:, 0], coords[:, 1]
        
        return i * (self._nsims - 1) + (j - 1)
```

micro_manager/adaptivity/global_adaptivity.py
```diff
@@ -85,12 +85,15 @@ def __init__(
         # Size of data type
         itemsize = MPI.FLOAT.Get_size()
 
+        # Number of similarity distances to be stored (strictly lower triangular)
+        n_dists = int(
+            self._global_number_of_sims * (self._global_number_of_sims - 1) / 2
+        )
```
Snapex2409, line 91: this is not fully right
if we assume strict tri mat of size n, then it has the same amount of entries of non strict tri mat of size n-1
meaning this should be (n-1) * (n-2) / 2

micro_manager/adaptivity/adaptivity.py
```diff
@@ -269,24 +278,19 @@ def _l1rel(self, data: np.ndarray) -> np.ndarray:
...
-        return np.linalg.norm(relative, ord=1, axis=-1)
-
-    def _l2rel(self, data: np.ndarray) -> np.ndarray:
+        for i in range(1, self._nsims):
+            for j in range(i):
+                p_diff = np.abs(data[i] - data[j])
+                denom = np.maximum(np.abs(data[i]), np.abs(data[j]))
+                rel_diff = p_diff / np.maximum(denom, eps)
+                flat_idx = self._map2dto1d([i, j])
+                self._similarity_dists[flat_idx] += np.linalg.norm(rel_diff, ord=1) * dt
```
Snapex2409: utilizing the proposed changes to the data and mapping the inner loop can be vectorized:

```suggestion
        for i in range(1, self._nsims):
            p_diff = np.abs(data[i] - data[i+1:self._nsims])
            denom = np.maximum(np.abs(data[i])[None, :], np.abs(data[i+1:self._nsims]))
            rel_diff = p_diff / np.maximum(denom, eps)
            base_idx = self._map2dto1d([i, i+1])
            self._similarity_dists[base_idx:base_idx+p_diff.shape[0]] += np.linalg.norm(rel_diff, ord=1) * dt
```

micro_manager/adaptivity/adaptivity.py
```diff
@@ -171,7 +174,10 @@ def _check_for_activation(self, inactive_id: int, active_ids: np.ndarray) -> boo
         tag : bool
             True if the inactive simulation needs to be activated, False otherwise.
         """
-        dists = self._similarity_dists[inactive_id, active_ids]
+        flat_ids = [
+            self._map2dto1d([inactive_id, active_id]) for active_id in active_ids
+        ]
+        dists = self._similarity_dists[flat_ids]
```
Snapex2409: vectorized:

```suggestion
        coords = np.zeros((len(active_ids), 2))
        coords[:, 0] = inactive_id
        coords[:, 1] = active_ids[:]
        flat_ids = self._map2dto1d(coords)
        dists = self._similarity_dists[flat_ids]
```

micro_manager/adaptivity/adaptivity.py
```diff
@@ -194,14 +200,15 @@ def _check_for_deactivation(self, active_id: int, active_ids: list) -> bool:
         """
         for active_id_2 in active_ids:
             if active_id != active_id_2:  # don't compare active sim to itself
+                flat_idx = self._map2dto1d([active_id, active_id_2])
                 # If active sim is similar to another active sim, deactivate it
-                if self._similarity_dists[active_id, active_id_2] < self._coarse_tol:
+                if self._similarity_dists[flat_idx] < self._coarse_tol:
                     return True
         return False
```
Snapex2409: vectorize:
```suggestion
        active_ids = np.array(active_ids, dtype=np.int32)
        if active_ids.shape[0] == 1 and active_ids[0] == active_id:
            return False
        mask = active_ids != active_id
        active_ids_to_check = active_ids[mask]
        coords = np.zeros((active_ids_to_check.shape[0], 2), dtype=np.int32)
        coords[:, 0] = active_id
        coords[:, 1] = active_ids_to_check[:]
        flat_idx = self._map2dto1d(coords)
        dists = self._similarity_dists[flat_idx]
        return np.any(dists < self._coarse_tol)
```

Snapex2409 (changes requested):
This should make lookup times faster. I cannot guarantee 100% correctness. But at least in my last sweep over the suggestions I did not see issues.

Hope this helps.

IshaanDesai:
Closing this PR because the size reduction of the similarity distance matrix leads to very high lookup costs, without reducing the N² complexity of the distance calculation. I will collect the lessons learned from this endeavor in an issue.

PR #241, opened 2026-04-06
Memory optimizations to norm calculations and a fix to lazy initialization

IshaanDesai:
This PR consists of two fixes:

- Avoid creation of intermediate arrays in norm calculations to save memory.
- Fix lazy initialization when there are ranks without any active simulations.

Checklist:

### DIFF
--- a/micro_manager/adaptivity/adaptivity.py
+++ b/micro_manager/adaptivity/adaptivity.py
@@ -275,16 +275,14 @@ class AdaptivityCalculator:
         similarity_dists : numpy array
             Updated 2D array having similarity distances between each micro simulation pair
         """
-        pointwise_diff = data[np.newaxis, :] - data[:, np.newaxis]
-        # divide by data to get relative difference
-        # divide i,j by max(abs(data[i]),abs(data[j])) to get relative difference
-        denom = np.maximum(
-            np.absolute(data[np.newaxis, :]), np.absolute(data[:, np.newaxis])
-        )
-        # Add small epsilon to avoid division by zero (invalid value warning) when both are 0
         eps = np.finfo(np.float64).eps
-        relative = pointwise_diff / np.maximum(denom, eps)
-        return np.linalg.norm(relative, ord=1, axis=-1)
+        data_abs = np.absolute(data)
+        denom = np.maximum(data_abs[np.newaxis, :], data_abs[:, np.newaxis])
+        return np.linalg.norm(
+            (data[np.newaxis, :] - data[:, np.newaxis]) / np.maximum(denom, eps),
+            ord=1,
+            axis=-1,
+        )
 
     def _l2rel(self, data: np.ndarray) -> np.ndarray:
         """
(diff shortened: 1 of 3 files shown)


micro_manager/adaptivity/adaptivity.py
```diff
@@ -275,16 +275,14 @@ def _l1rel(self, data: np.ndarray) -> np.ndarray:
...
         eps = np.finfo(np.float64).eps
-        relative = pointwise_diff / np.maximum(denom, eps)
-        return np.linalg.norm(relative, ord=1, axis=-1)
+        data_abs = np.absolute(data)
+        denom = np.maximum(data_abs[np.newaxis, :], data_abs[:, np.newaxis])
+        return np.linalg.norm(
+            (data[np.newaxis, :] - data[:, np.newaxis]) / np.maximum(denom, eps),
+            ord=1,
+            axis=-1,
+        )
```
Snapex2409:
```suggestion
        data_bc = data[np.newaxis, :]
        data_abs = np.absolute(data_bc)
        denom = np.maximum(data_abs, np.swapaxis(data_abs, 0, 1))
        return np.linalg.norm(
            (data_bc - np.swapaxis(data_bc, 0, 1) / np.maximum(denom, eps),
            ord=1,
            axis=-1,
        )
```

Snapex2409: To further reduce mem cost the following might work (not 100% sure about it)
would need to actually measure impact:
Idea is: np.swapaxis should return a view and not allocate memory. Thus, reducing the total amount of N² allocations.

Snapex2409: Tried to split comment and suggestion... but still cannot directly apply it.

micro_manager/adaptivity/adaptivity.py
```diff
@@ -301,13 +299,11 @@ def _l2rel(self, data: np.ndarray) -> np.ndarray:
...
         eps = np.finfo(np.float64).eps
-        relative = pointwise_diff / np.maximum(denom, eps)
-        return np.linalg.norm(relative, ord=2, axis=-1)
+        data_abs = np.absolute(data)
+        denom = np.maximum(data_abs[np.newaxis, :], data_abs[:, np.newaxis])
+        return np.linalg.norm(
+            (data[np.newaxis, :] - data[:, np.newaxis]) / np.maximum(denom, eps),
+            ord=2,
+            axis=-1,
+        )
```
Snapex2409: same as with l1rel

micro_manager/micro_manager.py
```diff
@@ -717,9 +721,9 @@ def initialize(self) -> None:
                 "The initialize() method of the Micro simulation requires initial data, but no initial macro data has been provided."
             )
 
-        initial_micro_data = None
+        initial_micro_data: dict[str, list] = dict()
 
-        if self._micro_sims_init:
+        if are_there_sims_to_init:
```
Snapex2409: It seems to me, that this will disregard whether micro sims actually implement initialize or not.
`self._micro_sims_init` contained that information.

Snapex2409: perhaps just add `and self._micro_sims_init`

micro_manager/micro_manager.py
```diff
@@ -811,14 +813,14 @@ def initialize(self) -> None:
...
+                initial_micro_data_list: list[dict] = [
                     dict(zip(initial_micro_data, t))
                     for t in zip(*initial_micro_data.values())
                 ]
             else:
                 # Ranks without active simulations provide empty dicts
-                initial_micro_data_list = [
+                initial_micro_data_list: list[dict] = [
                     dict() for _ in range(self._local_number_of_sims)
                 ]
```
Snapex2409, line 826: I think for the case in which the micro simulation class does not implement the `initialize` method, this will likely fail. Might need to fall back to using `initial_data` for this case.

Snapex2409 (commented):
I hope my comments make sense.

micro_manager/adaptivity/adaptivity.py
IshaanDesai: Let me test this out.

micro_manager/micro_manager.py
IshaanDesai: Yes indeed. This check only covers the case when lazy initialization is on, and, if so, whether the rank has active simulations. But applying a condition like:

```python
if self._micro_sims_init
```
seems not to work for ranks without any active problems.

PR #252, opened 2026-04-27
Fix model adaptivity infinite loop at resolution boundary

### LINKED ISSUE #251: Model adaptivity switching function loops endlessly
Let's assume a simple model adaptivity function:

```py
def switching_function(resolution, location, t, input, prev_output):
    return 1 if t > 0.5 else 0
```

This leads to an infinite loop as soon as t surpasses 0.5 as the function always return 1.

The correct switching function is:

```py
def switching_function(resolution, location, t, input, prev_output):
    return 1 if resolution == 0 and t > 0.5 else 0
```

Even though the switching function is wrong, it should not result in an infinite loop.
Possible solutions:

- Abort after some max-iterations have been reached.
- Ignore or raise an error for the wish to go to a finer resolution if already at the finest resolution.

Geethapranay1:
Fixes #251

This pr prevents the model-adaptivity loop from running forever when a switching function repeatedly requests a resolution change beyond the available model range

The convergence check now compares the current resolution against the clamped target resolution, so invalid boundary requests are treated as no-op changes.

This also fixes `switch_models()` to return a flat list of switched local ids. Previously, `np.argwhere(...).tolist()` returned nested lists such as [[0]], which could cause switched simulations to be missed when filtering cached outputs.

 added regression tests covering:

 - invalid switch request at finest/coarsest boundary,
 - valid switch detection,
 - manager loop exiting on invalid boundary requests,
 - actual model switch followed by repeated invalid request,
 - no stale output reuse after a model switch.

Also documented the boundary behavior in the model-adaptivity docs and added the new regression test to CI.
Checklist:

### DIFF
--- a/micro_manager/adaptivity/model_adaptivity.py
+++ b/micro_manager/adaptivity/model_adaptivity.py
@@ -219,7 +219,7 @@ class ModelAdaptivity:
             sims[idx].destroy()
             sims[idx] = sim_new
 
-        return np.argwhere((current_res - target_res) != 0).tolist()
+        return np.flatnonzero(current_res != target_res).tolist()
 
     def update_states(
         self,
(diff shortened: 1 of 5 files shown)


Geethapranay1:
@fsimonis PTAL

micro_manager/adaptivity/model_adaptivity.py
```diff
@@ -304,15 +304,10 @@ def check_convergence(
...
-            if active_sims[idx] != 1:
-                continue
-            prev_out = prev_output[idx] if prev_output is not None else None
-            next_switch[idx] = self._switching_func(
-                resolutions[idx], locations[idx], t, inputs[idx], prev_out
-            )
-        local_num_changes = np.sum(next_switch != 0)
+        target_resolutions = self._gather_target_resolutions(
+            resolutions, locations, t, inputs, prev_output, active_sims
+        )
```
IshaanDesai, line 309: @Snapex2409, do you see any potential pitfalls in gathering target resolutions all at once rather than individually?

tests/unit/test_model_adaptivity.py
IshaanDesai, line 1: Can you add docstrings to each test to clearly state what is being tested? I see that the test names are long, but docstrings are still valuable.

micro_manager/adaptivity/model_adaptivity.py
Snapex2409: Not really. It makes the code cleaner overall and handles the invalid resolution change request. All good from my side.

Snapex2409 (commented):
Seems good to me. Just need the test doc strings.

tests/unit/test_model_adaptivity.py
Geethapranay1: Sure, give me some time to write clearly then I will push the updated file.

PR #261, opened 2026-05-11
Fix the adaptivity data check

### LINKED ISSUE #260: Error when initialize() returns more data than specified for adaptivity
When the micro simulation returns more data than necessary by the adaptivity computation, then I see this error message:

```
The initialize() method needs to return data which is required for the adaptivity calculation.
```

This is misleading as the data is returned.

I would expect no error and only if data is actually missing, I would expect

```
The initialize() method needs to return data which is required for the adaptivity calculation. Missing: u, v.
```

fsimonis:
This PR changes the error message when the initialize() function of the Micro Simulations doesn't exactly return the data expected for adaptivity.

It now only errors when data is missing.
Having extra data leads to a warning.

Closes #260

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 ## latest
 
+- Allow `initialize()` to return data that is not used by the adaptivity [#261](https://github.com/precice/micro-manager/pull/261)
 - Fixed `MicroSimulation` initialization requiring positional parameters [#255](https://github.com/precice/micro-manager/pull/255)
 - Fixed model adaptivity convergence at resolution boundaries to prevent infinite loops for out-of-range switching requests [#252](https://github.com/precice/micro-manager/pull/252)
 - Add function `set_global_id` to the dummies and the example in the integration test [#247](https://github.com/precice/micro-manager/pull/247)
-- Allow `initialize()` to return data that is not used by the adaptivity [#261](https://github.com/precice/micro-manager/pull/261)
 
 ## v0.9.0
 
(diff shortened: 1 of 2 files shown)


micro_manager/micro_manager.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python3micro_manager
```
fsimonis:
```suggestion
#!/usr/bin/env python3
```

micro_manager/micro_manager.py
```diff
@@ -742,6 +742,20 @@ def initialize(self) -> None:
...
                 if self._is_adaptivity_on:
+                    # Check for missing data
+                    expected, provided = set(self._data_for_adaptivity.keys()), set(
+                        initial_micro_output.keys()
+                    )
+                    if missing := expected - provided:
+                        raise Exception(
+                            "The initialize() method needs to return data which is required for the adaptivity calculation. "
+                            f'Of the expected data {", ".join(expected)}, the following is missing: {", ".join(missing)}'
+                        )
```
IshaanDesai, line 753: This condition will always throw an Exception when macro-scale data is required in the adaptivity, because it will not be returned by the micro simulation. The check should throw an Exception when the provided value is not part of the expected data, rather than returning the same value.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Fixed `MicroSimulation` initialization requiring positional parameters [#255](https://github.com/precice/micro-manager/pull/255)
 - Fixed model adaptivity convergence at resolution boundaries to prevent infinite loops for out-of-range switching requests [#252](https://github.com/precice/micro-manager/pull/252)
 - Add function `set_global_id` to the dummies and the example in the integration test [#247](https://github.com/precice/micro-manager/pull/247)
+- Allow `initialize()` to return data that is not used by the adaptivity [#261](https://github.com/precice/micro-manager/pull/261)
```
IshaanDesai: Please move this entry to the top of the *latest* entries. The changelog follows a newest-first ordering.

IshaanDesai (changes requested):

micro_manager/micro_manager.py
fsimonis: I am using adaptivity_micro_data now as expected here. This shoulc be what you expect, right?

IshaanDesai: Works now :+1: I tested it with the solverdummy.

IshaanDesai (approved):

PR #262, opened 2026-05-11
Cleanup the import logic and improve errors

fsimonis:
This PR cleans up the import logic and improves import error message.

- The [import_module](https://docs.python.org/3/library/importlib.html#importlib.import_module) function was used incorrectly. This is now fixed. ~~I am still unsure why it works when passing a file with a `.py` suffix as it expects a module.~~ The location gets converted to a python module path after reading the config. So all good.
- The import module step is now a separate step and errors are split into 3 groups:
  - The configured module cannot be found
  - The configured module has dependencies that cannot be found
  - Something else went wrong
- Then the class is detected with a potential suffix of 0-9
- Finally the class is wrapped

Checklist:

### DIFF
--- a/micro_manager/micro_simulation.py
+++ b/micro_manager/micro_simulation.py
@@ -610,7 +610,7 @@ def load_backend_class(path_to_micro_file: str) -> type:
             )
         else:
             raise RuntimeError(
-                f"Counldn't load dependency of python module '{path_to_micro_file}' containing the micro simulation: {ie}"
+                f"Counld not load a dependency of the python module '{path_to_micro_file}' containing the micro simulation: {ie}"
             )
     except Exception as e:
         raise RuntimeError(
--- a/micro_manager/micro_simulation.py
+++ b/micro_manager/micro_simulation.py
@@ -606,7 +606,7 @@ def load_backend_class(path_to_micro_file: str) -> type:
     except ModuleNotFoundError as ie:
         if ie.name == path_to_micro_file:
             raise RuntimeError(
-                f"Couldn't load python module '{path_to_micro_file}' containing the micro simulation"
+                f"Could not load the python module '{path_to_micro_file}' containing the micro simulation"
             )
         else:
             raise RuntimeError(
--- a/micro_manager/micro_simulation.py
+++ b/micro_manager/micro_simulation.py
@@ -600,39 +600,46 @@ def load_backend_class(path_to_micro_file: str) -> type:
         A class inheriting from MicroSimulationInterface.
     """
 
-    def try_load(name):
-        try:
-            return getattr(ipl.import_module(path_to_micro_file, name), name)
-        except ImportError as ie:
-            return None
-        except AttributeError as ae:
-            return None
+    # try to load the module
+    try:
+        mod = ipl.import_module(path_to_micro_file)
+    except ModuleNotFoundError as ie:
+        if ie.name == path_to_micro_file:
+            raise RuntimeError(
+                f"Couldn't load python module '{path_to_micro_file}' containing the micro simulation"
+            )
+        else:
+            raise RuntimeError(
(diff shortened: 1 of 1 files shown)


micro_manager/micro_simulation.py
```diff
@@ -600,39 +600,46 @@ def load_backend_class(path_to_micro_file: str) -> type:
...
-        if not inherits:
-            cls = _wrap_non_interface_class(cls, path_to_micro_file)
-        return cls
+    # try to load the module
+    try:
+        mod = ipl.import_module(path_to_micro_file)
+    except ModuleNotFoundError as ie:
+        if ie.name == path_to_micro_file:
+            raise RuntimeError(
+                f"Couldn't load python module '{path_to_micro_file}' containing the micro simulation"
```
IshaanDesai:
```suggestion
                f"Could not load the python module '{path_to_micro_file}' containing the micro simulation"
```

micro_manager/micro_simulation.py
```diff
@@ -600,39 +600,46 @@ def load_backend_class(path_to_micro_file: str) -> type:
...
+            )
+        else:
+            raise RuntimeError(
+                f"Counldn't load dependency of python module '{path_to_micro_file}' containing the micro simulation: {ie}"
```
IshaanDesai:
```suggestion
                f"Could not load the dependency of the python module '{path_to_micro_file}' containing the micro simulation: {ie}"
```

IshaanDesai (commented):
Looks good from my side. @Snapex2409 can you please review because you last changes these files?

micro_manager/micro_simulation.py
fsimonis:
```suggestion
                f"Counld not load a dependency of the python module '{path_to_micro_file}' containing the micro simulation: {ie}"
```

micro_manager/micro_simulation.py
```diff
@@ -600,39 +600,46 @@ def load_backend_class(path_to_micro_file: str) -> type:
...
-        except TypeError:
-            # pybind11 extension types may not support issubclass — wrap them
-            inherits = False
-
-        if not inherits:
-            cls = _wrap_non_interface_class(cls, path_to_micro_file)
-        return cls
+    # try to load the module
+    try:
+        mod = ipl.import_module(path_to_micro_file)
```
fsimonis, line 605: I recently stumbled across this, which should be the way to go:

https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly

So either changing sys.path or calling runpy.run_path.

fsimonis: Ok. This isn't necessary since you assume a relative import from the execution location and convert the path to a module import.

This also explains why the current module import works.

micro_manager/micro_simulation.py
```diff
@@ -600,39 +600,46 @@ def load_backend_class(path_to_micro_file: str) -> type:
...
-        if result is not None:
-            return check_cls(result)
-
-    # failed to load any class
-    raise RuntimeError(f"Could not load micro simulation from {path_to_micro_file}")
+    suffixes = [""] + [str(i) for i in range(10)]
+    micro_cls = None
+    for suffix in suffixes:
+        if cls := getattr(mod, f"{CLS_NAME}{suffix}", None):
+            micro_cls = cls
```
Snapex2409, line 626:
```suggestion
            micro_cls = cls
            break
```
I might consider placing a break here, just in case to avoid unexpected behaviour.

Snapex2409 (commented):
Looks good to me, I just have one small comment:

micro_manager/micro_simulation.py
fsimonis: Good suggestion. Thanks!

IshaanDesai (approved):

PR #263, opened 2026-05-11
Fix OpenMPI cache miss in CI workflows

### LINKED ISSUE #259: Unit tests CI fails when launched via branch or pull request
The unit tests CI fails when launched via a branch merge or a pull request. An example of the failure: https://github.com/precice/micro-manager/actions/runs/25560350578/job/75030003225. When the failed job is manually triggered, it succeeds.

Geethapranay1:
This pr fixes CI failures on branch or pr runs by building OpenMPI when the cache is missing, instead of assuming ~/openmpi always exists

closes #259 

Checklist:

### DIFF
--- a/.github/workflows/build-openmpi-cache.yml
+++ b/.github/workflows/build-openmpi-cache.yml
@@ -1,0 +1,58 @@
+name: Build OpenMPI cache
+
+on:
+  push:
+    branches:
+      - main
+      - develop
+  pull_request:
+    branches:
+      - "*"
+  workflow_dispatch:
+
+jobs:
+  build_cache:
+    name: Build and cache OpenMPI 5
+    runs-on: ubuntu-latest
+    container: precice/precice:nightly
+    steps:
+      - name: Install build dependencies
+        run: |
+          apt-get -qq update
+          apt-get -qq install wget build-essential
+
+      - name: Load Cache OpenMPI 5
+        id: ompi-cache-load
+        uses: actions/cache/restore@v5
+        with:
+          path: ~/openmpi
+          key: openmpi-5.0.5-${{ runner.os }}-build
+
+      - name: Build OpenMPI 5
+        if: steps.ompi-cache-load.outputs.cache-hit != 'true'
+        run: |
+          set -euo pipefail
+          wget https://download.open-mpi.org/release/open-mpi/v5.0/openmpi-5.0.5.tar.gz
+          tar -xzf openmpi-5.0.5.tar.gz
+          cd openmpi-5.0.5
+          mkdir -p ~/openmpi
+          ./configure --prefix=$HOME/openmpi
+          make -j$(nproc)
+          make install
+
+      - name: Save OpenMPI 5 to cache
+        if: steps.ompi-cache-load.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v5
+        with:
+          path: ~/openmpi
+          key: openmpi-5.0.5-${{ runner.os }}-build
+
+      - name: Verify OpenMPI installation
+        run: |
+          if [ ! -d ~/openmpi ]; then
+            echo "::error::OpenMPI cache directory not found"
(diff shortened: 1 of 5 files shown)


Geethapranay1:
@IshaanDesai PTAL

https://github.com/precice/micro-manager/actions/runs/25684751288/job/75405373475

Snapex2409 (commented):
Thank you for the fix. After thinking about this for a bit, I think we should take this opportunity to address the underlying issue.

Even with this implementation, I think there is the possibility for race conditions to occur. 

Instead, I thought it might be good to create a single workflow file solely responsible for checking/building/storing the MPI cache. This WF would run when changes are pushed to a PR.
Then, all other workflows, that require the cache would ***not*** run as soon as changes are pushed to a PR branch, rather await the completion of the MPI cache workflow.
Towards this, we could make use of: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run

Regardless, the proposed solution is better than what we have. So, my proposal is open for discussion.

Geethapranay1:
hey @Snapex2409 , Thanks for the suggestion i agree, a centralized cache workflow is the cleaner approach as it eliminates the race conditions and keeps the build logic in one place.

One thing I noticed from the [workflow_run docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run) is that workflows triggered by workflow_run run with GITHUB_SHA and GITHUB_REF pointing to the default branch, not the PR branch. So the downstream test workflows would need to explicitly checkout the PR code using github.event.workflow_run.head_sha just wanted to flag this since it adds some extra checkout logic.

happy to implement this approach I'll update the pr.

Geethapranay1:
@Snapex2409 should i want to update this pr, or open a new pr

IshaanDesai:
@Geethapranay1 I suggest updating this PR only. It would be good if we could iterate on the idea by @Snapex2409 within this PR itself.

Geethapranay1:
Sure

Geethapranay1:
@Snapex2409 @IshaanDesai PTAL! i've updated the pr with the centralized cache workflow we talked about

- added build-openmpi-cache.yml to handle just the MPI caching.
- updated the downstream test workflows to use workflow_run and explicitly check out the pr code.
- synced with the latest develop branch and updated the changelog.

You'll notice the downstream tests (like unit tests and coverage) aren't running right now. This is a known GitHub Actions limitation workflow_run only triggers if the parent workflow already exists on the default branch (develop)

Since the cache workflow is brand new in this pr, the tests won't trigger right now. However, once this pr is merged, it will work perfectly for all future PRs. (If this is a dealbreaker, we could switch to a reusable workflow_call approach instead, but this current setup does exactly what we want once merged)

Let me know what you think

.github/workflows/build-openmpi-cache.yml
```diff
@@ -0,0 +1,58 @@
...
+        id: ompi-cache-load
+        uses: actions/cache/restore@v5
+        with:
+          path: ~/openmpi
+          key: openmpi-5.0.5-${{ runner.os }}-build
+
+      - name: Build OpenMPI 5
+        if: steps.ompi-cache-load.outputs.cache-hit != 'true'
+        run: |
+          set -euo pipefail
```
Snapex2409: Is the option -o needed here? Seems like this is not supported on the Runner Image.

Snapex2409 (commented):
Overall looks good to me. Thank you for your efforts.
Will need the final approval from @IshaanDesai.

.github/workflows/build-openmpi-cache.yml
Geethapranay1: Removed it  as github actions already runs bash steps with -eo pipefail by default, so it wasn't needed. Thanks for pointing it out

IshaanDesai (approved):
Looks good 👍

PR #264, opened 2026-05-15
Redesigned configuration handling

### LINKED ISSUE #154: Add defaults or require mandatory inputs for configuration parameters
Currently the JSON configuration input parameters do not have default values (at least some of them). The mandatory parameters need not have default values, but then must return exceptions when not set. A thorough overhaul of the configuration functionality is necessary.

Snapex2409:
This implementation should simplify future changes to configuration entries.
Error handling was shifted towards special loading contexts.

The core concepts are:

- There is no need for a long list of backing fields anymore. New config entry "getters" just need to be marked with `@config_entry`
-  Loading entries from file uses the `ConfigDSL`, which provides the methods `get_or_none`, `get_with_default` and `get_or_raise`. These methods can accept default args and format strings for success or failure.
- The internal config entry naming scheme was slightly altered: Since "getters" are now enriched with further attributes, the "get" prefix was dropped. Entries representing bool flags now begin with "enable" instead of "turn_on"

Checklist:

### DIFF
--- a/micro_manager/config.py
+++ b/micro_manager/config.py
@@ -17,6 +17,7 @@ class ConfigDSL:
     Provides a standardized context to read data from JSON.
     Data retrieval can be set up to yield optionals, with default values or raise Errors.
     """
+
     def __init__(self, data: Dict, log: Logger):
         """
         Constructs a context for the JSON data.
@@ -68,12 +69,12 @@ class ConfigDSL:
         return True
 
     def get_or_none(
-            self,
-            fmt_success: Optional[str]=None,
-            fmt_error: Optional[str]=None,
-            dtype: Optional[Type]=None,
-            options: Optional[List]=None,
-            **kwargs
+        self,
+        fmt_success: Optional[str] = None,
+        fmt_error: Optional[str] = None,
+        dtype: Optional[Type] = None,
+        options: Optional[List] = None,
+        **kwargs,
     ) -> Optional[Any]:
         """
         Resolves the value specified by the set path. Returns the result if available, otherwise None.
@@ -99,16 +100,18 @@ class ConfigDSL:
         result: Optional[Any]
             None if path not available or other failure, else value of JSON.
         """
-        return self.get_with_default(None, fmt_success, fmt_error, dtype, options, **kwargs)
+        return self.get_with_default(
+            None, fmt_success, fmt_error, dtype, options, **kwargs
+        )
 
     def get_with_default(
-            self,
-            default: Any,
-            fmt_success: Optional[str] = None,
-            fmt_error: Optional[str] = None,
-            dtype: Optional[Type] = None,
-            options: Optional[List] = None,
-            **kwargs
+        self,
+        default: Any,
+        fmt_success: Optional[str] = None,
+        fmt_error: Optional[str] = None,
+        dtype: Optional[Type] = None,
+        options: Optional[List] = None,
+        **kwargs,
     ) -> Any:
         """
         Resolves the value specified by the set path. Returns the result if available, otherwise the specified default.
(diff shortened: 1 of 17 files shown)


micro_manager/config.py
```diff
@@ -3,8 +3,433 @@
...
+    def get_or_raise(
+        self,
+        fmt_success: Optional[str] = None,
+        fmt_error: Optional[str] = None,
+        dtype: Optional[Type] = None,
+        options: Optional[List] = None,
+        **kwargs,
+    ):
+        """
+        Resolves the value specified by the set path. Returns the result if available, otherwise throws.
```
IshaanDesai, line 168: otherwise throws? Throws Error?

micro_manager/config.py
```diff
@@ -13,83 +438,35 @@ class Config:
...
-
-    def set_logger(self, logger):
+            Path to the JSON configuration file
+        """
+        self._config_file_name: str = config_file_name
+        self._base_dir: str = os.path.dirname(
+            os.path.join(os.getcwd(), config_file_name)
+        )
+        self._logger_null: Logger = Logger(Config.__name__, level=logging.CRITICAL)
+        self._logger: Logger = self._logger_null
```
IshaanDesai, line 455: Instead of this can we use the logging wrapper with a custom configuration logger within it?

micro_manager/config.py
```diff
@@ -100,504 +477,438 @@ def set_logger(self, logger):
...
+            ]["coarsening_constant"].get_or_raise(
+                "Adaptivity coarsening constant: {data}"
+            )
+            self.adaptivity_refining_constant.set = self.json["simulation_params"][
+                "adaptivity_settings"
+            ]["refining_constant"].get_or_raise("Adaptivity refining constant: {data}")
+            self.adaptivity_similarity_measure.set = self.json["simulation_params"][
+                "adaptivity_settings"
+            ]["similarity_measure"].get_with_default(
+                "L1",
```
IshaanDesai:
```suggestion
                "L2rel",
```

micro_manager/config.py
```diff
@@ -606,100 +917,92 @@ def read_json_micro_manager(self):
...
-                "in every time window."
-            )
+        # TODO what? this is not being saved nor used
+        diagnostics_data_names = self.json["diagnostics"][
+            "data_from_micro_sims"
+        ].get_or_none(
+            None,
+            "No diagnostics data is defined. Micro Manager will not output any diagnostics data.",
+            list,
+        )
```
IshaanDesai, line 941: I think this has no effect anymore, but I need to confirm once.

IshaanDesai (approved):
The proposed change to the configuration module makes a lot of sense. I just have a few questions before merging this.

micro_manager/config.py
Snapex2409: Yes, I was under the impression this was acceptable terminology.

micro_manager/config.py
Snapex2409: Sorry, I am not following. Could you elaborate?

micro_manager/config.py
IshaanDesai: Okay, now I understand what you mean.

micro_manager/config.py
IshaanDesai: To put it another way: why do we need to create a new Logger in `config.py`? Can't we simply use the logger acquired from the logging functionality?

Snapex2409: This newly created logger is not used to display any outputs. Its level is set to critical, so all messages will be ignored per the logging_wrapper API. This logger effectively redirects outputs to nowhere, which is useful for setting some default values for non-mandatory fields while not displaying error messages.

The real logger is still set using the `set_logger` method, which originates from `MicroManager`.

micro_manager/config.py
IshaanDesai: This is a separate issue. I will look at this individually later. Please leave the TODO in there so that I know this needs to be looked at.

IshaanDesai:
Ready to squash and merge. Please add a changelog, as this is a significant change.

PR #276, opened 2026-05-25
Fix load balancing case when there is a mismatch in send and receive problems

IshaanDesai:
When a non-zero threshold value is used, the number of micro simulations to send and receive may differ. In such cases, the load balancing should simply check whether either the number of simulations to send or receive is zero, and if so, not perform balancing. The condition to check this was incorrect, leading to a failure. This PR fixes that.

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## latest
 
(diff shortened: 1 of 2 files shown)


fsimonis (approved):
This resolves my failing cases 👍

PR #273, opened 2026-05-20
Minor fixes in MicroManager regarding Adaptivity

Snapex2409:
Some variable in the `MicroManager` were named incorrectly.
Switched `gid` to `lid`.

Checklist:

- [x] I made sure that the CI passed before I ask for a review.
- [x] I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`.
- [x] If necessary, I made changes to the documentation and/or added new content.
- [x] I will remember to squash-and-merge, providing a useful summary of the changes of this PR.

PR #271, opened 2026-05-20
Export model resolution when using Model Adaptivity

Snapex2409:
Export `model_resolution` when Model Adaptivity is enabled.

Checklist:

- [x] I made sure that the CI passed before I ask for a review.
- [x] I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`.
- [x] If necessary, I made changes to the documentation and/or added new content.
- [x] I will remember to squash-and-merge, providing a useful summary of the changes of this PR.

PR #255, opened 2026-05-05
Fix parameter inspection for dynamic MicroSimulation class

Snapex2409:
As encountered in issue [#248](https://github.com/precice/micro-manager/issues/248), the initialize method
of the dynamically created `CompatibilityWrapper` class passes parameters as `*args, **kwargs`.
Subsequent inspection using `inspect.getfullargspec` does not recognize that positional parameters are required.
This results in no input data being provided to the initialize method.

The provided implementation remedies that by exposing positional parameters.

Checklist:

- [x] I made sure that the CI passed before I ask for a review.
- [x] I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`.
- [x] If necessary, I made changes to the documentation and/or added new content.
- [x] I will remember to squash-and-merge, providing a useful summary of the changes of this PR.

PR #280, opened 2026-05-26
Release v0.10.0

IshaanDesai:
This release consists of:

- Fixed load balancing for case where number of sims to send and recv are not the same [#276](https://github.com/precice/micro-manager/pull/276)
- Fixed CI OpenMPI cache miss by centralizing the MPI build into a dedicated workflow [#263](https://github.com/precice/micro-manager/pull/263)
- Fixed naming of variables tracking active simulations in the `MicroManager` [#273](https://github.com/precice/micro-manager/pull/273)
- Improved load balancing of inactive simulations by tracking a small amount of time [#272](https://github.com/precice/micro-manager/pull/272)
- Added the exported field `model_resolution` when using model adaptivity [#271](https://github.com/precice/micro-manager/pull/271)
- Fixed model adaptivity active simulation mask generation for 0 local simulations [#270](https://github.com/precice/micro-manager/pull/270)
- Allow `initialize()` to return data that is not used by the adaptivity [#261](https://github.com/precice/micro-manager/pull/261)
- Fixed `MicroSimulation` initialization requiring positional parameters [#255](https://github.com/precice/micro-manager/pull/255)
- Fixed model adaptivity convergence at resolution boundaries to prevent infinite loops for out-of-range switching requests [#252](https://github.com/precice/micro-manager/pull/252)
- Add function `set_global_id` to the dummies and the example in the integration test [#247](https://github.com/precice/micro-manager/pull/247)

### DIFF
--- a/docs/model-adaptivity.md
+++ b/docs/model-adaptivity.md
@@ -169,3 +169,5 @@ The output is expected to be an integer and is interpreted in the following mann
 | 0     | No resolution change                                  |
 | -1    | Increase model fidelity by one (go back one in list)  |
 | 1     | Decrease model fidelity by one (go one ahead in list) |
+
+If the switching function requests a change beyond the available resolution range, the request is ignored and does not trigger another model-adaptivity iteration.
--- a/.github/workflows/build-openmpi-cache.yml
+++ b/.github/workflows/build-openmpi-cache.yml
@@ -1,0 +1,57 @@
+name: Build OpenMPI cache
+
+on:
+  push:
+    branches:
+      - main
+      - develop
+  pull_request:
+    branches:
+      - "*"
+  workflow_dispatch:
+
+jobs:
+  build_cache:
+    name: Build and cache OpenMPI 5
+    runs-on: ubuntu-latest
+    container: precice/precice:nightly
+    steps:
+      - name: Install build dependencies
+        run: |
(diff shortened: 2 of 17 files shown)


docs/model-adaptivity.md
```diff
@@ -169,3 +169,5 @@ The output is expected to be an integer and is interpreted in the following mann
 | 0     | No resolution change                                  |
 | -1    | Increase model fidelity by one (go back one in list)  |
 | 1     | Decrease model fidelity by one (go one ahead in list) |
+
+If the switching function requests a change beyond the available resolution range, the request is ignored and does not trigger another model-adaptivity iteration.
```
Snapex2409, line 173:
```suggestion
If the switching function requests a change beyond the available resolution range, the request is clamped within the available range. Further requests beyond the range are ignored and do not trigger another model-adaptivity iteration.
```

Snapex2409 (commented):
Overall, looks good to me. Only found one minor inconsistency in the documentation.

docs/model-adaptivity.md
IshaanDesai: Thanks! I'll change this on develop and it will be reflected in a future release.

MakisH (approved):
A quick sanity check looks good, I have not looked into technical details.

PR #281, opened 2026-05-27
Fix parameter inspection of initialize() when micro simulation is pybind11 wrapped

IshaanDesai:
When a pybind11-based micro simulation is provided, `inspect.getfullargspec()` on `initialize()` returns `TypeError: unsupported callable`. This PR now handles this error and, if the' TypeError' is hit, directly passes whatever input arguments are available without inspection. The ideal solution would be to find a way to inspect arguments of a pybind11-based micro simulation. This is a temporary fix.

Checklist:

### DIFF
--- a/micro_manager/micro_simulation.py
+++ b/micro_manager/micro_simulation.py
@@ -555,16 +555,21 @@ def __getattr__(self, name):
     # Only add initialize override if the wrapped class actually has it,
     # so that requires_initialize() returns True for those classes.
     if has_initialize:
-        argspec = inspect.getfullargspec(cls.initialize)
-        # build args
-        init_args = f"{', '.join(argspec.args)}"
-        params = f"{', '.join(argspec.args[1::])}"
-        if argspec.varargs is not None:
-            init_args += f", *args"
-            params += f", *args"
-        if argspec.varkw is not None:
-            init_args += f", **kwargs"
-            params += f", **kwargs"
+        try:
+            argspec = inspect.getfullargspec(cls.initialize)
+            # build args
+            init_args = f"{', '.join(argspec.args)}"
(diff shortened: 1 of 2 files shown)


micro_manager/micro_simulation.py
```diff
@@ -555,16 +555,21 @@ def __getattr__(self, name):
...
+            if argspec.varargs is not None:
+                init_args += f", *args"
+                params += f", *args"
+            if argspec.varkw is not None:
+                init_args += f", **kwargs"
+                params += f", **kwargs"
+        except TypeError:
+            # pybind11 methods do not expose Python-style argument specs
+            init_args = "self, *args, **kwargs"
+            params = "*args, **kwargs"
```
Snapex2409:
```suggestion
        try:
            argspec = inspect.getfullargspec(cls.initialize)
            # build args
            params = f"{', '.join(argspec.args[1:])}"
            if argspec.varargs is not None:
                params += f", *args"
            if argspec.varkw is not None:
                params += f", **kwargs"
        except TypeError:
            # pybind11 methods do not expose Python-style argument specs
            has_param = False
            try:
                test_instance = cls(-1)
                test_instance.initialize()
            except TypeError:
                # calling without args failed, need to provide one
                has_param = True
            
            pos_arg = "input_data, " if has_param else ""
            params = f"{pos_arg}*args, **kwargs"
        init_args = f"self, {params}"
```

Snapex2409 (commented):
This should fix the case in which simulations require data for the initialize method.

IshaanDesai:
> This should fix the case in which simulations require data for the initialize method.

The fix works for a case where `initialize` requires input data. Thanks a lot.

PR #282, opened 2026-05-27
Fix passing of macro mesh coordinates when model adaptivity is run in parallel

IshaanDesai:
When model adaptivity is run, the switching function requires the coordinate location. Even if this is run in parallel, only the local coordinates are required. This PR simplifies passing the coordinates.

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Micro Manager changelog
 
+## latest
+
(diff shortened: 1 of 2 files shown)


Snapex2409 (commented):
If my understanding is correct, this will yield incorrect behaviour.
The proposed changes will certainly result in non-crashing simulations, but now wrong macro coords will be provided.

`mesh_vertex_coords` are in the cases of load balancing or global adaptivity the coordinates across the entire domain. In the other cases, it only contains the coords of the localized access region.

PR #284, opened 2026-05-28
Switch to dedicated simulation container

Snapex2409:
The implementation provides a dedicated simulation container that encapsulates all micro-simulation data-related aspects.
This should lay the groundwork for follow-up encapsulation refactorings.

The proposed container keeps track of local GIDs, simulation instances, and corresponding macro coordinates, and provides additional utility methods. Adding and removing simulations can now be performed by calling the respective methods. Assigning simulation instances can be performed by using the index operator.

Checklist:

### DIFF
--- a/micro_manager/simulation_container.py
+++ b/micro_manager/simulation_container.py
@@ -5,6 +5,14 @@ import numpy as np
 
 
 class SimulationContainer:
+    """
+    The simulation container stores all simulation instances of this rank alongside
+    the corresponding GIDs and macro-scale coordinates. Further, this container manages
+    checkpointing by providing methods to store and load simulation states. Lastly, it
+    provides functionality to insert and remove simulation instances as well as further
+    utility methods and properties.
+    """
+
     EntryType = Tuple[
         Optional[MicroSimulationInterface],
         Dict[str, Optional[Any]],
@@ -14,7 +22,8 @@ class SimulationContainer:
 
     def __init__(self):
         """
-        Constructs SimContainer. When model adaptivity is active, state storage needs to capture states of all models.
+        Constructs SimulationContainer.
+        When model adaptivity is active, state storage must capture the states of all models.
 
         Parameters
         ----------
@@ -192,7 +201,7 @@ class SimulationContainer:
         for lid in self.range_lid:
             self._sim_checkpoints[lid].clear()
 
-    def get_state(self, lid: int) -> Dict[str, Any]:
+    def get_sim_state(self, lid: int) -> Dict[str, Any]:
         """
         Gets the state of the simulation at the given local id.
 
@@ -215,7 +224,7 @@ class SimulationContainer:
 
         return state_dict
 
-    def set_state(self, lid: int, state: Dict[str, Any]) -> None:
+    def set_sim_state(self, lid: int, state: Dict[str, Any]) -> None:
         """
         Sets the state of the simulation at the given local id to the provided state.
 
@@ -237,7 +246,7 @@ class SimulationContainer:
 
     def __len__(self) -> int:
         """
-        Returns the number of local simulations.
+        Returns the number of simulations on this rank (local).
 
         Returns
         -------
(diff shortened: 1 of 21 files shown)


micro_manager/simulation_container.py
```diff
@@ -0,0 +1,355 @@
...
+    EntryType = Tuple[
+        Optional[MicroSimulationInterface],
+        Dict[str, Optional[Any]],
+        int,
+        np.ndarray,
+    ]
+
+    def __init__(self):
+        """
+        Constructs SimContainer. When model adaptivity is active, state storage needs to capture states of all models.
```
IshaanDesai:
```suggestion
        Constructs SimulationContainer. When model adaptivity is active, state storage must capture the states of all models.
```
Can you please add a few sentences on what the simulation container is? Something that, if someone reads for the first time, gives an idea of the simulation container.

micro_manager/simulation_container.py
```diff
@@ -0,0 +1,355 @@
...
+            sim.set_state(state_dict[key])
+
+    def clear_checkpoints(self) -> None:
+        """
+        Deletes all stored checkpoints.
+        """
+        for lid in self.range_lid:
+            self._sim_checkpoints[lid].clear()
+
+    def get_state(self, lid: int) -> Dict[str, Any]:
```
IshaanDesai:
```suggestion
    def get_sim_state(self, lid: int) -> Dict[str, Any]:
```
I recommend renaming this to `get_sim_state` to avoid confusion with the MicroSimulationInterface function `get_state`.

micro_manager/simulation_container.py
```diff
@@ -0,0 +1,355 @@
...
+        sim = self._sims[lid]
+        assert sim is not None
+
+        state_dict = self._sim_checkpoints[lid]
+        key = f"{sim.name}-state"
+        state_dict[key] = sim.get_state()
+
+        return state_dict
+
+    def set_state(self, lid: int, state: Dict[str, Any]) -> None:
```
IshaanDesai:
```suggestion
    def set_sim_state(self, lid: int, state: Dict[str, Any]) -> None:
```
Same reason as stated before.

micro_manager/simulation_container.py
```diff
@@ -0,0 +1,355 @@
...
+
+        sim = self._sims[lid]
+        assert sim is not None
+        key = f"{sim.name}-state"
+
+        sim.set_state(state_dict[key])
+
+    def __len__(self) -> int:
+        """
+        Returns the number of local simulations.
```
IshaanDesai:
```suggestion
        Returns the number of simulations on this rank (local).
```

micro_manager/simulation_container.py
```diff
@@ -0,0 +1,355 @@
...
+        -------
+        num_sims : int
+            Number of local simulations.
+        """
+        return len(self._sims)
+
+    def __iter__(self) -> Iterable[EntryType]:
+        """
+        Iterates over all entries in this container. Each entry is a tuple of a potential
+        simulation instance, its checkpoint data, the GID and spatial coordinate.
```
IshaanDesai:
```suggestion
        simulation instance, its checkpoint data, the GID, and the macro-scale coordinate.
```

IshaanDesai (commented):
I looked at all the changes. Thanks for the effort, it looks great already! I just have a few suggestions in the `SimulationContainer` class. Additionally, I noticed that a call to `get_global_id` was removed from model adaptivity. Does this mean that we can get rid of the micro simulation functions `get_global_id` and `set_global_id`?

micro_manager/simulation_container.py
Snapex2409: Added it to the class doc string.

Snapex2409:
Regarding the micro simulation methods.
No, we cannot.
They are still in use by the Tasking Model / Workers (needed to determine the storage location).

IshaanDesai (approved):
Looks good! Please squash and merge.

PR #285, opened 2026-05-29
Rename tolerance to distinguish from time in doc

Fujikawas:
't'  is used for both `tolerance` and `time` in these equations, which can be confusing.

Checklist:


IshaanDesai (commented):
Rather than saying `tol`, it would be better to use a mathematical symbol. It would be worth looking at what the typical mathematical symbols are for tolerances.

PR #287, opened 2026-06-01
Refactor P2P and MPI handling

Snapex2409:
This PR provides an `MPIHandler` that manages an MPI communicator and provides utility functions for it. All P2P logic was moved to this new class and abstracted. Specifically, the `p2p_comm` function was generalised and is now named `exchange`. This method is used across Load Balancing, Adaptivity, and Interpolation as the p2p comm backbone, rather than using their individual implementations.

Checklist:


micro_manager/adaptivity/global_adaptivity.py
```diff
@@ -555,3 +515,17 @@ def _update_inactive_sims(self) -> None:
         self._precice_participant.stop_last_profiling_section()
 
         self._sim_is_associated_to = np.copy(_sim_is_associated_to_updated)
+
+    def _create_comm_map(
```
IshaanDesai:
```suggestion
    def _create_send_map(
```

micro_manager/adaptivity/global_adaptivity.py
```diff
@@ -519,18 +484,13 @@ def _update_inactive_sims(self) -> None:
...
                 )
 
-        recv_reqs = p2p_comm(
-            self._sim_container.local_gids,
-            self._rank,
-            self._comm,
-            self._sim_container.global_num_sims,
-            list(to_be_activated_map.keys()),
-            sim_states_and_global_ids,
+        send_map = self._create_comm_map(
```
IshaanDesai:
```suggestion
        send_map = self._create_send_map(
```

micro_manager/tools/mpi_handler.py
```diff
@@ -0,0 +1,224 @@
+from typing import Optional, Tuple, List, Any, Dict, Union
+
```
IshaanDesai:
```suggestion
```

micro_manager/tools/mpi_handler.py
```diff
@@ -0,0 +1,224 @@
...
+        return self._size
+
+    @property
+    def rank(self) -> int:
+        return self._rank
+
+    def is_parallel(self) -> bool:
+        return self.size > 1
+
+    def create_node_buffer(self, dtype: MPI.Datatype, count: int):
```
IshaanDesai, line 33:
```suggestion
    def create_matrix_node_buffer(self, dtype: MPI.Datatype, count: int):
```
As we create a buffer of size `count * count * itemsize`, it is clear that it is for a matrix. Let's explicitly state that in the function name.

micro_manager/tools/mpi_handler.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        Constructs a numpy array buffer of the given dimensions for the ranks located on the same node.
```
IshaanDesai, line 35:
```suggestion
        Constructs a 2D symmetric numpy array buffer of the given dimensions for the ranks located on the same node.
```

micro_manager/tools/mpi_handler.py
```diff
@@ -0,0 +1,224 @@
...
+
+        return obj_to_rank
+
+    def create_empty_exchange_map(self) -> Dict[int, List[Any]]:
+        return {r: [] for r in range(self.size)}
+
+    def exchange(
+        self,
+        send_map: Dict[int, List[Any]],
+        /,
```
IshaanDesai, line 112: Why is this line necessary?

IshaanDesai (commented):
The refactoring makes a lot of sense. Thanks for doing this! I have only minor comments. The core MPI functionality itself remains the same.

micro_manager/tools/mpi_handler.py
Snapex2409: Thank you for catching that. The function itself should create a buffer of size `count * itemsize`, not a matrix.
I will remove the second occurrence of count, thus keeping the method general.
In GlobalAdaptivity, where it is used, the provided count will be of the size of the desired matrix.

micro_manager/tools/mpi_handler.py
Snapex2409: I added it, such that one can only provide the other parameters as keyword args.
The intention was to make it very explicit when one wants to deviate from default behaviour, such that it does not happen by accident.

IshaanDesai: Got it.

micro_manager/tools/mpi_handler.py
IshaanDesai: Okay. As the method is now general, the suggestion no longer applies.

IshaanDesai (approved):
Ready to squash and merge.

PR #293, opened 2026-06-08
Added missing documentation for workers

Snapex2409:
Added the missing documentation for workers.

Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -269,8 +269,8 @@ micro simulation computation.
 With `backend`, the communication between the Micro Manager rank and its workers can be set to use either a
 socket-based or MPI-based approach. For a SLURM system, by setting `is_slurm` to true, the worker process is
 attempted to be started with `srun` instead of `mpiexec`. In this case, `backend` must be set to `"socket"`.
-`num_workers` controls the amount of workers per Micro Manager rank. When `num_workers` is less or equal than `1`, then
-local execution is assumed, thus no worker processes are created.
+`num_workers` controls the amount of workers per Micro Manager rank. When `num_workers` is `1`, then
+local execution is assumed, thus no worker processes are created. Only values greater `0` are accepted.
 `mpi_impl` should be set to the underlying MPI implementation, as this is required for pinning.
 
 | Parameter       | Description                                            | Default    |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -263,11 +263,11 @@ The following parameters can be set
 ## Tasking / Workers
 
 See the [tasking](tooling-micro-manager-nested-parallelization.html) documentation for a detailed explanation of the implementation.
-The usage of workers can be enabled by configuring the tasking parameters to values other than the default.
+The usage of workers can be enabled by configuring the tasking parameters.
 When using workers, a separate MPI process group is started per Micro Manager rank. This should allow for parallel
 micro simulation computation.
 With `backend`, the communication between the Micro Manager rank and its workers can be set to use either a
-socket based or MPI based approach. For system with SLURM: By setting `is_slurm` to true the worker process is
+socket-based or MPI-based approach. For a SLURM system, by setting `is_slurm` to true, the worker process is
 attempted to be started with `srun` instead of `mpiexec`. In this case, `backend` must be set to `"socket"`.
 `num_workers` controls the amount of workers per Micro Manager rank. When `num_workers` is less or equal than `1`, then
 local execution is assumed, thus no worker processes are created.
@@ -280,7 +280,7 @@ local execution is assumed, thus no worker processes are created.
 | `"num_workers"` | Number of workers pre Micro Manager rank.              | `1`        |
 | `"mpi_impl"`    | Implementation type of MPI. Options: ["intel", "open"] | `"open"`   |
 
-The following configuration block should be provided on the same level as the `"simulation_param"`.
+The following configuration block should be provided on the same level as the simulation parameters.
 The given example uses socket based communication, launches with `mpiexec`, uses 4 workers per rank and
 assumes Intel MPI.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -260,6 +260,38 @@ The following parameters can be set
 }
 ```
 
+## Tasking / Workers
+
+The usage of workers can be enabled by configuring the tasking parameters to values other than the default.
+When using workers, a separate MPI process group is started per Micro Manager rank. This should allow for parallel
+micro simulation computation.
+With `backend`, the communication between the Micro Manager rank and its workers can be set to use either a
+socket based or MPI based approach. For system with SLURM: By setting `is_slurm` to true the worker process is
+attempted to be started with `srun` instead of `mpiexec`. In this case, `backend` must be set to `"socket"`.
+`num_workers` controls the amount of workers per Micro Manager rank. When `num_workers` is less or equal than `1`, then
+local execution is assumed, thus no worker processes are created.
+`mpi_impl` should be set to the underlying MPI implementation, as this is required for pinning.
+
+| Parameter       | Description                                            | Default    |
+|-----------------|--------------------------------------------------------|------------|
+| `backend`       | Communication backend. Options: ["socket", "mpi"]      | `"socket"` |
+| `is_slurm`      | Launch worker with `srun`?                             | `"False"`  |
+| `"num_workers"` | Number of workers pre Micro Manager rank.              | `1`        |
+| `"mpi_impl"`    | Implementation type of MPI. Options: ["intel", "open"] | `"open"`   |
+
+The following configuration block should be provided on the same level as the `"simulation_param"`.
+The given example uses socket based communication, launches with `mpiexec`, uses 4 workers per rank and
+assumes Intel MPI.
+
+```json
+"tasking": {
+        "backend": "socket",
+        "is_slurm": false,
+        "num_workers": 4,
+        "mpi_impl": "intel"
+    }
+```
+
 ## Interpolate a crashed micro simulation
 
 If the optional dependency `sklearn` is installed, the Micro Manager will derive the output of a crashed micro simulation by interpolating outputs from similar simulations. To enable this, set
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -262,6 +262,7 @@ The following parameters can be set
 
 ## Tasking / Workers
 
+See the [tasking](tooling-micro-manager-nested-parallelization.html) documentation for a detailed explanation of the implementation.
 The usage of workers can be enabled by configuring the tasking parameters to values other than the default.
 When using workers, a separate MPI process group is started per Micro Manager rank. This should allow for parallel
 micro simulation computation.
--- a/docs/nested-parallelization.md
+++ b/docs/nested-parallelization.md
@@ -1,28 +1,28 @@
 ---
 title: Running with multi-level parallelization
-permalink: tooling-micro-manager-nested-parallelization.html
-keywords: tooling, macro-micro, parallelization, workers, tasking
-summary: Micro Manager supports MPI parallelized micro solvers with nested parallelization.
+permalink: tooling-micro-manager-multi-level-parallelization.html
+keywords: tooling, macro-micro, parallelization, multi-level parallelization, workers, tasking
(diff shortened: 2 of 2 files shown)


docs/configuration.md
```diff
@@ -260,6 +260,39 @@ The following parameters can be set
 }
 ```
 
+## Tasking / Workers
+
+See the [tasking](tooling-micro-manager-nested-parallelization.html) documentation for a detailed explanation of the implementation.
+The usage of workers can be enabled by configuring the tasking parameters to values other than the default.
```
IshaanDesai: Is this true, though? Ideally we should have a boolean to enable tasking and then the user can optionally set tasking parameters. I guess what you want to say here is: *The usage of workers can be enabled by configuring the tasking parameters.* and not values other than defaults.

docs/configuration.md
```diff
@@ -260,6 +260,39 @@ The following parameters can be set
...
+When using workers, a separate MPI process group is started per Micro Manager rank. This should allow for parallel
+micro simulation computation.
+With `backend`, the communication between the Micro Manager rank and its workers can be set to use either a
+socket based or MPI based approach. For system with SLURM: By setting `is_slurm` to true the worker process is
```
IshaanDesai:
```suggestion
socket-based or MPI-based approach. For a SLURM system, by setting `is_slurm` to true, the worker process is
```

docs/configuration.md
```diff
@@ -260,6 +260,39 @@ The following parameters can be set
...
+attempted to be started with `srun` instead of `mpiexec`. In this case, `backend` must be set to `"socket"`.
+`num_workers` controls the amount of workers per Micro Manager rank. When `num_workers` is less or equal than `1`, then
```
IshaanDesai: `num_workers` less than 1 makes no sense, right? We should throw an error if the value is less than 1.

docs/configuration.md
```diff
@@ -260,6 +260,39 @@ The following parameters can be set
...
+`mpi_impl` should be set to the underlying MPI implementation, as this is required for pinning.
+
+| Parameter       | Description                                            | Default    |
+|-----------------|--------------------------------------------------------|------------|
+| `backend`       | Communication backend. Options: ["socket", "mpi"]      | `"socket"` |
+| `is_slurm`      | Launch worker with `srun`?                             | `"False"`  |
+| `"num_workers"` | Number of workers pre Micro Manager rank.              | `1`        |
+| `"mpi_impl"`    | Implementation type of MPI. Options: ["intel", "open"] | `"open"`   |
+
+The following configuration block should be provided on the same level as the `"simulation_param"`.
```
IshaanDesai:
```suggestion
The following configuration block should be provided on the same level as the simulation parameters.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
+---
+title: Running with multi-level parallelization
+permalink: tooling-micro-manager-nested-parallelization.html
+keywords: tooling, macro-micro, parallelization, workers, tasking
+summary: Micro Manager supports MPI parallelized micro solvers with nested parallelization.
```
IshaanDesai:
```suggestion
title: Running with multi-level parallelization
permalink: tooling-micro-manager-multi-level-parallelization.html
keywords: tooling, macro-micro, multi-level parallelization, workers, tasking
summary: Micro Manager supports MPI parallelized micro solvers with multi-level task parallelization.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+---
+
+## Overview
+
+With default configurations, running the Micro Manager in parallel with MPI results in $N$ ranks within the same MPI communicator.
```
IshaanDesai:
```suggestion
By default, running the Micro Manager in parallel with MPI results in $N$ ranks within the same MPI communicator.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+In this approach, each operates under the assumption, that Micro Simulations are either solved in serial or use shared memory parallelization.
```
IshaanDesai:
```suggestion
In this approach, each Micro Manager instance assumes that micro simulations are either solved in serial or use shared-memory parallelization.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+If one attempts to solve the Micro Simulation with MPI, this will result in incorrect behavior, as different Micro Simulation instances communicate with each other. Further, this will likely lead to race conditions or deadlocks.
```
IshaanDesai:
```suggestion
If one attempts to run individual micro simulations with MPI, incorrect behavior occurs because different micro simulation instances communicate with each other. Further, this will likely lead to race conditions or deadlocks.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+
+To support MPI parallelization for Micro Simulations, multi-level parallelization is introduced. In this execution mode, each Micro Manager rank $n\in [N]$ creates its own sub-process group of $M$ ranks with MPI.
```
IshaanDesai:
```suggestion
To support MPI parallelization for individual micro simulations, multi-level parallelization is introduced. In this execution mode, each Micro Manager rank $n\in [N]$ creates its own sub-process group of $M$ ranks with MPI.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+This results in a total of $N\cdot M$ processes. The sub-process groups are launched to run a "worker" process, and will be henceforth called workers. The worker executable is a generic task handler. Its sole task is to receive incoming tasks, decode them, solve the request and send back the results. Workers have a persistent storage, to which all incoming tasks have access.
```
IshaanDesai:
```suggestion
This results in a total of $N\cdot M$ processes. Each subprocess group consisting of $M$ processes launches a "worker" process. The subprocess groups are henceforth referred to as workers. The worker executable is a generic task handler. It receives incoming tasks, decodes them, solves the request, and sends back the results. Workers have a persistent storage, to which all incoming tasks have access.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+
+The Micro Manager makes use of this, by first initializing the persistent storage with Micro Manager specific task-classes and loading the relevant Micro Simulation classes.
+When the Micro Manager attempts to solve a Micro Simulation, this results in a "Solve-Request" being emitted and transferred to its $M$ workers.
+Each of those, receive the request and solve the specified simulation in parallel. Upon completion, the results are transferred back to the Micro Manager rank. Other operations within the Micro Simulation Interface are handled similarly.
```
IshaanDesai:
```suggestion
The Micro Manager deploys this parallelization by first initializing the persistent storage with Micro Manager-specific task classes and loading the relevant micro simulation classes.
Solving a micro simulation results in a "Solve-Request" being emitted and transferred to its $M$ workers.
Each worker receives the request and solves the specified simulation in parallel. Upon completion, the results are transferred back to the Micro Manager rank. All micro simulation calls are parallelized in this manner.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Tasks
+
+In accordance with the Micro Simulation Interface, each interface method has its corresponding Task implementation.
```
IshaanDesai: Can you add a relative link to the `class MicroSimulation` documentation?

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+During operation no task objects are pickled (serialized) and transferred between the Micro Manager and its workers. Instead, only a string representation of the respective task class and the required data is transferred.
+This is done for performance reasons. The task classes are only transferred once during worker initialization and are stored under the key "tasks" in each workers persistent storage.
```
IshaanDesai:
```suggestion
During operation, no task objects are pickled (serialized) and transferred between the Micro Manager and its workers. Instead, only a string representation of the respective task class and the required data is transferred.
This is done for performance reasons. The task classes are transferred only once during worker initialization and stored under the key "tasks" in each worker's persistent storage.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+state_data["tasks"] = task_dict
+state_data["sim_classes"] = dict()
+state_data["sim_instances"] = dict()
+state_data["load_function"] = load_function
+```
+
+The key `tasks` contains all possible task classes, that may be encountered during simulation.
+`sim_classes` is a dictionary mapping class names to loaded Micro Simulation classes.
+`sim_instances` is a dictionary mapping Micro Simulation GIDs to their respective Micro Simulation instance.
+`load_function` contains a function capable of loading a Micro Simulation class.
```
IshaanDesai:
```suggestion
The key `tasks` contains all possible task classes that may be encountered during simulation.
`sim_classes` is a dictionary mapping class names to loaded Micro Simulation classes.
`sim_instances` is a dictionary mapping Micro Simulation GIDs to their respective Micro Simulation instance.
`load_function` contains a function capable of loading a Micro Simulation class.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+    try:
+        task_descriptor = conn.recv(src_id)
+        output = handle_task(state_data, task_descriptor)
+        conn.send(dst_id, output)
+    except Exception:
+        break
+conn.close()
+```
+
+After receiving a `task_descriptor`, that was created by `Task.send_args`, `handle_task` creates an instance of the task and attempts to handle the request.
```
IshaanDesai:
```suggestion
After receiving a `task_descriptor` created by `Task.send_args`, `handle_task` creates a task instance and attempts to handle the request.
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+The resulting output is then sent back to the Micro Manager parent rank.
+
+## Connection
+
+Connections between workers and Micro Manager ranks are realized as either MPI or socket based communication.
+The MPI version (`MPIConnection`) calls internally `MPI_Comm_spawn`, which launches a separate process group and yields a MPI communicator between the Micro Manager parent and the newly create process group.
+This approach is known to be more error-prone and required the definition of the following environment variables:
```
IshaanDesai:
```suggestion
Connections between workers and Micro Manager ranks are realized as either MPI or socket-based communication.
The MPI version (`MPIConnection`) internally calls `MPI_Comm_spawn`, which launches a separate process group and yields an MPI communicator between the Micro Manager parent and the newly created process group.
This approach is known to be more error-prone and requires the definition of the following environment variables:
```

docs/nested-parallelization.md
```diff
@@ -0,0 +1,135 @@
...
+
+```bash
+export OMPI_MCA_btl=self,vader,tcp
+export OMPI_MCA_btl_tcp_if_include=lo
+export OMPI_MCA_oob_tcp_if_include=lo
```
IshaanDesai, line 108: Is there an MPI documentation reference that we could point to here?

IshaanDesai (approved):
Thank you for adding elaborate documentation. I have some suggestions. Please squash and merge once the suggestions are incorporated.

docs/configuration.md
Snapex2409: > Is this true?

As clarification: Factually, the sentence was correct. If the user does not provide any _tasking parameters_, the default number of workers is 1. This results in no workers being spawned.

However, my wording was certainly not ideal. Will use your suggestion.

docs/configuration.md
Snapex2409: When configuring the tasking parameters, the user might try to disable workers by specifying `num_workers` as 0 or 1.

This was my reasoning regarding <= 1.

docs/nested-parallelization.md
Snapex2409: In the [OpenMPI doc](https://www.open-mpi.org/faq/?category=tcp) those keywords are explained a bit.
But this exact combination I reached after a series of trial and error using hints from various forums.

docs/configuration.md
IshaanDesai: To avoid confusion and unintended behavior, I suggest throwing an error if `num_workers` is < 1. We can change the documentation here and fix the error in a separate PR.

PR #299, opened 2026-06-27
Refactored memory usage analysis to general diagnostics

Snapex2409:
Added a general diagnostics module that can be extended for further analysis.
Currently, only the memory usage has been refactored to follow this scheme.

__NOTE__: This PR should be merged after #298 

Checklist:

- [x] I made sure that the CI passed before I ask for a review.
- [x] I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`.
- [x] If necessary, I made changes to the documentation and/or added new content.
- [x] I will remember to squash-and-merge, providing a useful summary of the changes of this PR.

PR #300, opened 2026-06-27
Refactored LoadBalancing into an interface

Snapex2409:
Refactored LoadBalancing into an interface to reduce confusion for ActiveBalancer.
Moved counters from MicroManager to LoadBalancer implementations.

__NOTE__: Merge this PR after #299 

Checklist:

### DIFF
--- a/micro_manager/load_balancing.py
+++ b/micro_manager/load_balancing.py
@@ -820,7 +820,7 @@ class ActiveBalancer(LoadBalancer):
                 self._log.log_warning_rank_zero(
                     "It appears that the micro simulations are already fairly balanced. No load balancing will be done. Try changing the threshold value to induce load balancing."
                 )
-                return send_map, recv_map
+                return self._mpi.create_empty_exchange_map()
             if n_global_send_sims != 0 or n_global_recv_sims != 0:
                 ActiveBalancer._correct_active_exchange_data(
                     global_send_sims,
--- a/micro_manager/load_balancing.py
+++ b/micro_manager/load_balancing.py
@@ -1,10 +1,9 @@
 import time
+from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import Optional, Dict, List, Any, Set
 
 import numpy as np
-
-from precice import Participant
 
 from micro_manager.config import Config
 from micro_manager.adaptivity.adaptivity_interface import AdaptivityInterface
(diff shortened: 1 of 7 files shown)


micro_manager/load_balancing.py
```diff
@@ -31,89 +30,353 @@ class LoadBalancerSimData:
...
+        gid : int
+            Simulation GID
+        """
+        pass
+
+    @abstractmethod
+    def update(self):
+        """
+        Needs to be called after all micro simulations have finished their solve method.
+        Updates the load balancing metric and shares it globally.
```
IshaanDesai: It would be worth adding a note here stating that this is only valid for solve-time measurement-based load balancing.

micro_manager/load_balancing.py
```diff
@@ -31,89 +30,353 @@ class LoadBalancerSimData:
...
         """
-        self._enabled = config.enable_load_balancing()
+        self._log: Logger = log
+        self._mpi: MPIHandler = mpi
+        self._model_manager: ModelManager = model_manager
+        self._sim_container: SimulationContainer = sim_container
+        self._profiler: Profiler = None
+        self._adaptivity_controller: AdaptivityInterface = None
+
+    def initialize(self, profiler: Profiler, adaptivity: AdaptivityInterface):
```
IshaanDesai, line 62: Why is `initialize` required here?

micro_manager/load_balancing.py
```diff
@@ -31,89 +30,353 @@ class LoadBalancerSimData:
...
+
+        Parameters
+        ----------
+        sim_outputs: List[Dict[str, Any]]
+            Rank local simulation outputs.
+        """
+        for lid in self._sim_container.range_lid:
+            sim_outputs[lid]["rank_of_sim"] = self._mpi.rank
+
+    def _gather_send_data(
```
IshaanDesai, line 141: Is it correct that this function is only used when inactive simulations are also to be balanced by moving them to the ranks of their associated active counterparts? If so, lets mention this explicitly in the docstring.

micro_manager/load_balancing.py
```diff
@@ -31,89 +30,353 @@ class LoadBalancerSimData:
...
+    def _get_global_inactive_gids(
+        self, active_gids: Optional[Set[int]] = None
+    ) -> Set[int]:
+        """
+        Get global IDs of all inactive gids. This is based on local ids.
+
+        Parameters
+        ----------
+        active_gids: Set[int]
+            set of global active gids, if omitted will be recomputed.
```
IshaanDesai:
```suggestion
            set of global active gids, if omitted, will be recomputed.
```

IshaanDesai (commented):
Thanks for this. I have only minor suggestions and a few questions to clarify. I will now run a load-balancing regression test on this PR to ensure the underlying functionality has not changed.

micro_manager/load_balancing.py
Snapex2409: For load balancing in general:

initialize was added to be able to have a typed field in the micro manager (without any warnings)

Why is a part in the interface:

To ensure `_profiler` and `_adaptivity_controller` are populated for implementations.

micro_manager/load_balancing.py
Snapex2409: No, this method is used to collect data for all balancing activities.
Modified the doc-string slightly to clarify.

IshaanDesai: Much more understandable now :+1: thanks

IshaanDesai (approved):
Tested it with a regression test of the two-scale heat conduction.

PR #302, opened 2026-06-29
Prepare documentation for rendering in Hugo

MuhammadAashirAslam:
Preserve the existing micro-manager documentation URL for the Hugo migration

Checklist:

- [x] I made sure that the CI passed before I ask for a review.
- I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`. -> Not needed
- [x] If necessary, I made changes to the documentation and/or added new content.
- [x] I will remember to squash-and-merge, providing a useful summary of the changes of this PR.

PR #304, opened 2026-07-03
Change markdownlint pre-commit hook

MakisH:
In PR https://github.com/precice/micro-manager/pull/302, pre-commit is failing, complaining about spaces in lists, while the list that it is complaining about is in the frontmatter. Updating the respective pre-commit hook does not help, and excluding the rule to work around this issue would not be well-motivated.

This PR changes the respective pre-commit hook to https://github.com/igorshubovych/markdownlint-cli, used by other preCICE repositories.

It also fixes a few valid issues in some markdown files and removes rule exceptions that are not needed (currently not violated, and it would also be good to not violate them, to be consistent with the rest of the documentation).

It also updates the base pre-commit-hooks hook.

I tried updating black as well, but any newer version changes the formatting of the code, which I did not want to dive into at the moment.

Checklist:

- I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`. -> Not needed.
- If necessary, I made changes to the documentation and/or added new content. -> Not needed


IshaanDesai (approved):
Thanks for the updates!

Issue #305, opened 2026-07-04
Using adaptivity leads to only one active problem after the first time window


IshaanDesai:
Using the current `develop` branch, if local or global adaptivity is deployed, there is only one active problem after the first time window, no matter what the settings are. The `adaptivity-metrics-global.csv` for a two-scale heat conduction problem looks like:

```csv
n|n active|n inactive|avg active|avg inactive|max active|rank of max active|max inactive|rank of max inactive
0|1|199|0.5|99.5|1|1|129|1
5|1|199|0.5|99.5|1|1|129|1
10|1|199|0.5|99.5|1|1|129|1
15|1|199|0.5|99.5|1|1|129|1
20|1|199|0.5|99.5|1|1|129|1
25|1|199|0.5|99.5|1|1|129|1
30|1|199|0.5|99.5|1|1|129|1
35|1|199|0.5|99.5|1|1|129|1
40|1|199|0.5|99.5|1|1|129|1
45|1|199|0.5|99.5|1|1|129|1
50|1|199|0.5|99.5|1|1|129|1
```

Using v0.10.1 for the same setup leads to:

```csv
n|n active|n inactive|avg active|avg inactive|max active|rank of max active|max inactive|rank of max inactive
0|1|199|0.5|99.5|1|1|129|1
5|8|192|4.0|96.0|7|0|129|1
10|10|190|5.0|95.0|9|0|129|1
15|10|190|5.0|95.0|9|0|129|1
20|11|189|5.5|94.5|10|0|129|1
25|11|189|5.5|94.5|10|0|129|1
30|12|188|6.0|94.0|10|0|128|1
35|12|188|6.0|94.0|10|0|128|1
40|13|187|6.5|93.5|11|0|128|1
45|13|187|6.5|93.5|11|0|128|1
50|13|187|6.5|93.5|11|0|128|1
```

The result with the develop branch is clearly incorrect.

IshaanDesai:
Fixed in https://github.com/precice/micro-manager/pull/306

PR #306, opened 2026-07-04
Fix read data buffer updating and MPI irecv buffer size

Snapex2409:
As mentioned in issue #305, in the current version, the number of active simulations does not change after initialisation.

The underlying issue is that adaptivity provides a `read_buffer` to the `CouplingHandler`, which should be populated with data read from preCICE. However, due to the original handling of `None` values, such empty buffers were always ignored. Instead, a method-local buffer was used. This left the buffer provided by adaptivity empty. Hence, the adaptivity data buffers were not updated, which resulted in the constant number of active simulations.

PS:
Also contains a fix for the MPIHandler. Ran into the issue with insufficient buffer size again.
Added back the 1MB large temporary buffer for MPI `irecv`.

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## latest
 
-- Fixed macro data not populating adaptivity buffers [#306](https://github.com/precice/micro-manager/pull/306)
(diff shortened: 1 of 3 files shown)


IshaanDesai (approved):
Ran regression tests with the two-scale heat conduction problem and the tests pass 👍

PR #308, opened 2026-07-06
Fixed micro simulation activation issue

Snapex2409:
Before, all `to be activated` micro simulations were instantiated. However, only the subset with associated simulations on the same rank received their state via `set_state`. The subsequent check for `None` simulations failed to remove the remaining simulations. This led to `get_state` being called before `set_state` or `initialize`.

Checklist:

- [x] If this is a use-side change, I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`. Furthermore, if this ia a breaking change, I wrote **breaking change** at the end of the changelog entry.
- [x] I will remember to squash-and-merge and provide a useful summary of the changes of this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## latest
 
+- Fixed simulation activation issue of calling `get_state` prior to full initialization [#308](https://github.com/precice/micro-manager/pull/308)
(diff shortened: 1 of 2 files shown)


IshaanDesai (approved):
Thanks! Please merge.

PR #311, opened 2026-07-14
Fixed stateless wrapper

Snapex2409:
The stateless wrapper was broken during refactoring. 
No instances could be created as not all abstract methods were implemented.

Checklist:

- [x] If this is a use-side change, I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`. If this is a breaking change, I wrote **breaking change** at the end of the changelog entry.
- [x] I will remember to squash-and-merge and provide a useful summary of the changes of this PR.


IshaanDesai (approved):
Would it be worth adding a test?

PR #312, opened 2026-07-22
Minor fixes for load balancing and stateless simulations 

Snapex2409:
### Load Balancing:

Added an implicit coupling iteration counter to determine the frequency of load balancing.

### ModelManager:

Fixed the `ModelWrapper` for stateless simulations. Without the name property, it would return an incorrect class name during lookups. 

### Interpolation:

Fixed RBF basis functions. (brackets were placed incorrectly)
Fixed RBF interpolation domain selection. (mixed up tree height and depth)
Fixed interpolation normalization. Normalization scaling is now based on a fraction of the input and not constant.

### Adaptivity:

Fixed data size extraction of micro simulation input and output during adaptivity interpolation for empty micro manager ranks. 

Checklist:

- [ ] If this is a use-side change, I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`. If this is a breaking change, I wrote **breaking change** at the end of the changelog entry.
- [ ] I will remember to squash-and-merge and provide a useful summary of the changes of this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## latest
 
-- Various fixes for interpolation, load balancing and adaptivity. Load balancing now balances every N implicit iterations. [#312](https://github.com/precice/micro-manager/pull/312)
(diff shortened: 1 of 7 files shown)


IshaanDesai (approved):
Please don't forget to add a changelog entry as these fixes affect users.

PR #320, opened 2026-08-10
Normalize generated preCICE data names

### LINKED ISSUE #317: Naming convention of data added to the preCICE configuration is inconsistent
When an adaptive solution strategy is used, the scalar data `Active-State` and `Active-Steps` need to be added to the preCICE XML configuration file, as stated in the [documentation](https://precice.org/tooling-micro-manager-configuration.html#adding-adaptivity-in-the-precice-xml-configuration). Similarly, when load balancing is used, the data `rank_of_sim` needs to be added. When model switching is used, the data `model_resolution` should be added. The data names are inconsistent. Recommendation is to:

- change `rank_of_sim` to `Rank-Of-Sim`.
- change `model_resolution` to `Model-Resolution`.

alloutflo:
## Summary

- rename the generated load-balancing scalar from `rank_of_sim` to `Rank-Of-Sim`
- rename the generated model-adaptivity scalar from `model_resolution` to `Model-Resolution`
- keep configuration registration and runtime output keys in sync
- document both required preCICE data names and record the breaking change
- add regression coverage for configuration, load-balancing output, and model-adaptivity output

Fixes #317

## Root cause

The two optional outputs were registered in the Micro Manager configuration and emitted at runtime under snake_case names, while the other generated preCICE data names use the documented title-case, hyphenated convention. Changing only one side would make the configured write-data name disagree with the actual output dictionary, so this updates both paths together.

## Validation

- `python -m unittest test_micro_manager.py test_model_adaptivity.py test_load_balancing.py` — 18 tests passed, 6 parallel-only tests skipped in the serial run
- `mpiexec -n 2 python -m unittest test_load_balancing.py` — passed on both ranks
- `mpiexec -n 4 --oversubscribe python -m unittest test_load_balancing.py` — passed on all four ranks
- `pre-commit run --all-files --verbose` — all hooks passed
- `git diff --check` — clean

Checklist:

- [x] If this is a use-side change, I added a summary of the changes (compared to the last release) in the `CHANGELOG.md`. If this is a breaking change, I wrote **breaking change** at the end of the changelog entry.
- [ ] I will remember to squash-and-merge and provide a useful summary of the changes of this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Micro Manager changelog
 
+## Unreleased
+
+- Renamed the load-balancing output `rank_of_sim` to `Rank-Of-Sim` and the model-adaptivity output `model_resolution` to `Model-Resolution` for consistent preCICE data names. Existing preCICE configurations using either output must be updated. [#317](https://github.com/precice/micro-manager/issues/317) **Breaking change**
+
 ## v0.11.2
 
 - Fixed use of provided HDF5 output file name when snapshots are computed in parallel. [#315](https://github.com/precice/micro-manager/pull/315)
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -262,6 +262,11 @@ Example of model adaptivity configuration is
 }
 ```
 
+Model adaptivity also writes the scalar data `Model-Resolution` for every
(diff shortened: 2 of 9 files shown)


IshaanDesai (approved):
Thanks for this 👍

